PR #4, opened 2016-08-28
Fix DropBox's drop

sinkuu:
`drop(a)` does nothing because `a` is `&mut A`.


alexcrichton:
Ah oops, thanks for the sharp eye!

PR #13, opened 2016-09-03
Reorganize the entire crate:

### LINKED ISSUE #3: Rename `Loop`
The term `Loop` is a bit vague for role played by the type. The `Loop` type listens for event notifications, then dispatches the event notifications to their targets. It is a `Loop` in that it repeats this job until it is shutdown.

Three options discussed are:

**IoLoop**

Pro: It references I/O, which is the primary role of the component.
Con: It does more than I/O, so `IoLoop` doesn't really match the scope of the component.

**EventLoop**

Pro: Better describes that it is doing and the term has been used in other async I/O libraries, so there is prior art for it.
Con: `EventLoopHandle` is long

**Reactor**

Pro: Not too long and describes what the component is actually doing. There is also plenty of prior art for the term "reactor"
Con: The term comes with some baggage in that `Reactor` already has meaning to some people that isn't necessarily related to what our component does.

### LINKED ISSUE #6: Consider moving functions off of `LoopHandle` & `LoopPin`
I will use connecting a `TcpStream` as an example, but the idea applies to a number of other functions on `LoopHandle`.

The current primary API for connecting a TCP stream is to use `LoopHandle::tcp_connect` or `LoopPin::tcp_connect` (which doesn't currently exist, but I believe the current plan is to implement it). There also exists `TcpStream::connect_stream` which allows for more advanced socket configuration.

IMO there are two drawbacks to this strategy.

1) It creates a scattered API. When a user wants to discover how to establish a TCP stream, they will most likely first look at `TcpStream`. They will see `TcpStream::connect_stream` and one conclusion is that this is the way to connect a `TcpStream`. They will have to know to look at either `LoopHandle` or `LoopPin` depending on their case.

2) It is not an extensible pattern. The goal of tokio-core is to allow any `mio::Evented` type to work with `tokio_core::Loop`. Imagine `struct TimerFd`, a wrapper around Linux's timerfd feature. This library would not be able to easily implement `LoopHandle::create_timer()`, and will most likely fallback on the `TimerFd::create(&loop_handle)` API.

I propose to keep the functions on `LoopHandle` minimal in favor of keeping constructor functions on the types themselves:

``` rust
struct TcpStream {
    fn connect(addr: &SocketAddr, event_loop: &LoopHandle) -> IoFuture<TcpStream> {
        // ...
    }
}
```

This leaves open the question of how to differentiate between `LoopHandle`, which must return an `IoFuture` and `LoopPin` which can complete the operation immediately.

To handle this, two functions are needed:

``` rust
struct TcpStream {
    fn connect(addr: &SocketAddr, event_loop: &LoopPin) -> io::Result<TcpStream> {
        // ...
    }

    fn connect_on(addr: &SocketAddr, event_loop: &LoopHandle) -> IoFuture<TcpStream> {
        // ...
    }
}
```

However, `LoopPin` refers to the currently running `Loop` (this is due to the fact that it cannot be sent). So, one possible simplification would be to omit `LoopPin` in the first function in favor of having a way to retrieve it from a thread local:

``` rust
struct TcpStream {
    fn connect(addr: &SocketAddr) -> io::Result<TcpStream> {
        // ...
    }

    fn connect_on(addr: &SocketAddr, event_loop: &LoopHandle) -> IoFuture<TcpStream> {
        // ...
    }
}
```

In this case, if `TcpStream::connect` is called off of an event loop, the function would either return an error or it could always return `NotReady` until it is operated on from within the context of a running `Loop`.

alexcrichton:
Renamed APIs
- Loop => reactor::Core
- LoopHandle => reactor::Handle
- LoopPin => reactor::Pinned
- TcpStream => net::TcpStream
- TcpListener => net::TcpListener
- UdpSocket => net::UdpSocket
- Sender => channel::Sender
- Receiver => channel::Receiver
- Timeout => reactor::Timeout
- ReadinessStream => reactor::PollEvented
- All `LoopHandle` methods to construct objects are now free functions on the
  associated types, e.g. `LoopHandle::tcp_listen` is now `TcpListener::bind`
- All APIs taking a `Handle` now take a `Handle` as the last argument
- All future-returning APIs now return concrete types instead of trait objects

Added APIs
- io::Io trait -- Read + Write + ability to poll

Removed without replacement:
- AddSource
- AddTimeout
- IoToken
- TimeoutToken

Closes #3 
Closes #6


chrivers:
"All LoopHandle methods to construct objects are now free functions on the associated types, e.g. LoopHandle::tcp_listen is now TcpListener::bind". This is a great change. It was kind of confusing that LoopHandle seemingly had Tcp-specific code.

And the grouping of concepts into {reactor, net, channel} is very good. Much more logical separation, and should be easier to get into now, since the terms also more closely match those from other languages and models.

If I might make a plea, I think that (and others on the tokio gitter have expressed the same sentiment) Pin is a pretty confusing name. I guess the reason is that it's a "Pin"ned loop handle? (pinned to a thread, I means?)

Maybe something like ThreadTask? Or FixedHandle?

Pin is.. well it's certainly short. It's got that going for it :D

alexcrichton:
@chrivers yeah we were debating about `Pin` as well as it didn't seem great. Unfortunately we couldn't think of another better name and settled on `Pinned` which was at least a little longer and less generic. In general though it's conveying that it's "pinned to the event loop thread" as opposed to "handle" which is just an arbitrary handle to an event loop

PR #24, opened 2016-09-09
Add `turn` on `Core` to allow single event loop iterations

aidanhs:
This isn't particularly polished, I wanted to get some initial feedback.

Thoughts/notes: (edit: the below are mostly out of date with the reworked implementation)

~~1. it seemed to be more fitting to have `Core` be a `Stream`, because it's basically an infinite iterator, albeit one that can be altered in the middle of iteration (i.e. polling).
2. however, `Core` never actually produces anything, it just returns whether a loop iteration has been successful
3. the piece of code shown at [1](left unchanged) could probably be improved to return an Err when polling, rather than panicking
4. it only makes sense to call `poll` if you've previously called `spawn`
5. it's important that you can call `poll` on `Core` outside of a task, because otherwise you can't spawn new jobs into it~~

[1]

```
err @ Err(_) => {
    err.unwrap();
}
```

### DIFF
--- a/src/reactor/mod.rs
+++ b/src/reactor/mod.rs
@@ -5,6 +5,7 @@
 //! futures, schedule tasks, issue I/O requests, etc.
 
 use std::cell::RefCell;
+use std::cmp;
 use std::io::{self, ErrorKind};
 use std::mem;
 use std::rc::{Rc, Weak};
@@ -226,6 +227,18 @@ impl Core {
         res.expect("run should not return until future is done")
     }
 
+    /// Performs one iteration of the event loop, blocking on waiting for events
+    /// for at most `max_wait` (forever if `None`).
+    ///
+    /// It only makes sense to call this method if you've previously spawned
+    /// a future onto this event loop.
+    ///
+    /// `loop { lp.turn(None) }` is equivalent to calling `run` with an
+    /// empty future (one that never finishes).
+    pub fn turn(&mut self, max_wait: Option<Duration>) {
+        self.poll_internal(max_wait, &mut || true).expect("Error in event loop turn");
+    }
+
     fn _run(&mut self, done: &mut FnMut() -> bool) {
         // Check to see if we're done immediately, if so we shouldn't do any
         // work.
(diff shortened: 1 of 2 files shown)


aidanhs:
On reflection, I'm not sure implementing `Stream` is the correct abstraction here. One might be tempted to try and `Merge` it with another stream and end up with a busy-wait loop. This isn't necessarily wrong, but people should probably explicitly opt into it with `loop { core.poll_core() }` or similar.

This might also be appropriate because it doesn't really make sense for a reactor core to return NotReady.

aidanhs:
I've reworked this a bit. Rather than forcing the event loop into the abstraction of a stream, a new method `turn` (I don't really mind what it's called) is provided. `turn` permits bubbling up of errors encountered while waiting for events, in contrast to `run`.

I've also added some docs.

Edit: one thing I'm not 100% on is whether turn is ok to spuriously wake up on EINTR.

alexcrichton:
The implementation looks pretty reasonable, thanks! If this is being added for https://github.com/alexcrichton/futures-rs/issues/136 first, though, perhaps we should get to the bottom of that first? In isolation it's not immediately clear to me how this method would be leveraged in the context of a larger application perhaps.

alexcrichton:
Note that another implementation strategy here would to encode `core.run(futures::finished(()))` as "turn the event loop at least once"

aidanhs:
> You don't know when to come back to the reactor. If futures aren't ready you don't know when they'll be ready and otherwise just can't correctly proceed.

Some things that you might want to poll may not have notification mechanisms. If this is the case then you're fine with not knowing when you can proceed, you just poll once per $small_timeslice. Notification isn't required for correct behaviour (well, unless you have something that specifically requires you to use notifications), but it is a significant optimisation.

> We could force run(futures::done(Ok(())) to turn the event loop, being the equivalent of turn.

That works for a timeout of 0, but doesn't let you block with timeout. It'd be useful to either be able to specify a timeout on one turn, or have a `run_until` method which takes a duration. You can build the latter with the former, so I'm inclined for the former. I could be swayed though.

alexcrichton:
True yeah if we want a way to say "poll for at most X milliseconds for an event" there's no way to do that. It sounds like though that it could be encoded via the [shutdown mechanism](https://github.com/tokio-rs/tokio-core/issues/33) we're thinking of, however, perhaps?

aidanhs:
Hmm, I'm not quite sure how that helps. `turn` is designed to allow you to use a `Core` alongside other non-futures polling mechanisms (or other event loops, assuming they also implement `turn`), with the timeouts effectively allowing you to set a minimum frequency of your overarching `loop { ... }`.

Shutdown seems to provide a way to watch out for a point where there are no tasks 'in-flight' on the loop whereas turn intentionally returns control to the caller while tasks may still be 'in-flight' on the loop.

alexcrichton:
Ah sorry for being a bit late to get back to this! Thinking again I think you're right, and I'm fine going ahead and adding `turn`. Perhaps though we could remove the `io::Result` from the return signature as that's not returned by `run`, which is this somewhat mirroring as well?

aidanhs:
Removed the `io::Result`.

I initially wasn't convinced (what if you wanted to handle the errors?) but then I looked at the possible errors from `epoll_wait`:

```
       EBADF  epfd is not a valid file descriptor.
       EFAULT The memory area pointed to by events is not accessible with write permissions.
       EINTR  The call was interrupted by a signal handler before either (1) any of the requested events occurred or (2) the timeout expired; see signal(7).
       EINVAL epfd is not an epoll file descriptor, or maxevents is less than or equal to zero.
```

There's very little a user can do with anything except `EINTR`, and for that they just need to try again which they should be doing regularly with `turn` anyway. So I guess panicking is fine.

src/reactor/mod.rs
```diff
@@ -226,6 +227,18 @@ impl Core {
...
+    /// Performs one iteration of the event loop, blocking on waiting for events
+    /// for at most `max_wait` (forever if `None`).
+    ///
+    /// It only makes sense to call this method if you've previously spawned
+    /// a future onto this event loop.
+    ///
+    /// `loop { lp.turn(None) }` is equivalent to calling `run` with an
+    /// empty future (one that never finishes).
+    pub fn turn(&mut self, max_wait: Option<Duration>) {
+        self.poll_internal(max_wait, &mut || true).expect("Error in event loop turn");
```
alexcrichton, line 239: Could this call `_run` if we ensured that the loop was turned at least once even if the future is ready?

aidanhs: Yes, but I'm not clear on the best way to do that. The duration argument would obviously need adding to `_run`, but how should `turn` indicate to `_run` that it must loop at least once? A `bool` argument?

Note that it requires a bit of care, because if the timeout expires in `poll_internal` without an event then finished will be false (`done` will never get called) and `_run` will faithfully loop again.

alexcrichton:
Ok, I did a few touchups locally, but merged now. Thanks!

aidanhs:
Fantastic, thanks!

PR #28, opened 2016-09-11
proxy example and changes motivated by it

dwrensha:
I wanted to implement a simple proxy in the style of the [echo example](https://github.com/tokio-rs/tokio-core/blob/master/examples/echo.rs), but I ran into a problem: currently there's no way to `shutdown()` the write end of a `split()` socket.  A naive proxy that ignores this problem will fail to close connections properly.

So this pull request includes two changes to enable a simple proxy example:
1.  When complete, `io::copy()` returns the reader and writer, in addition to the number of bytes forwarded (following in the style of `io::read_exact()` and `io::write_all()`).
2. `ReadHalf` and `WriteHalf` gain `shutdown()` methods.


examples/proxy.rs
```diff
@@ -0,0 +1,58 @@
...
+        let amounts = server.and_then(move |server| {
+            let (client_read, client_write) = client.split();
+            let (server_read, server_write) = server.split();
+
+            let from_client = copy(client_read, server_write)
+                .and_then(|(_, mut server_write, n)| {
+                    server_write.shutdown().map(|()| n)
+                });
+
+            let from_server = copy(server_read, client_write).map(|(_, _, n)| n);
```
alexcrichton: Technically `shutdown()` should be propagated here as well, right?

examples/proxy.rs
```diff
@@ -0,0 +1,58 @@
...
+    let server_addr = env::args().nth(2).unwrap_or("127.0.0.1:8080".to_string());
+    let server_addr = server_addr.parse::<SocketAddr>().unwrap();
+
+    // Create the event loop that will drive this server.
+    let mut l = Core::new().unwrap();
+    let handle = l.handle();
+
+    // Create a TCP listener which will listen for incoming connections.
+    let socket = TcpListener::bind(&listen_addr, &l.handle()).unwrap();
+    println!("Listening on: {}", listen_addr);
```
alexcrichton, line 29: Perhaps this could also print out "proxying to: {}"

alexcrichton:
Seems slick to me! I agree that `Copy` should probably yield the I/O objects back to the caller of the future to do things like shutdown.

I wonder though if perhaps `ReadHalf` and `WriteHalf` could provide some form of accessor rather than hardcoding `shutdown` on `ReadHalf<TcpStream>`? Due to the `RefCell` it may need to be a `with` method rather than a `get_ref` method, but that may not be so bad in the grand scheme of things.

dwrensha:
> I wonder though if perhaps `ReadHalf` and `WriteHalf` could provide some form of accessor rather than hardcoding shutdown on `ReadHalf<TcpStream>`?

I actually did explore that approach. It led to a confusing bug where I accidentally called `TcpStream::shutdown(Write)` on the `ReadHalf` of the client socket rather than on the `WriteHalf` of the server socket. More generally, allowing `ReadHalf<T>` and `WriteHalf<T>` to obtain references to the underlying `T` seems to go against the spirit of having them be two distinct types. If you prefer having a `with()` method over my current approach, then maybe we should take that one step further and adjust `split()` so that it returns two values of the same type. It might even make sense to make that returned type implement `Clone`.

Or maybe there should be `TcpStream::split()` where the returned values automatically call `shutdown()` on being dropped?

examples/proxy.rs
dwrensha: Yep. Good point.

dwrensha:
Or maybe the `Io` trait should have a `shutdown()` method which we could call from impls of `Drop` for `ReadHalf<T>` and `WriteHalf<T>`.

alexcrichton:
Ah yeah I could see mixing these up definitely. My worry is just that lots of times we won't have literally `SplitRead<TcpStream>` but rather something like `SplitRead<TlsStream<TcpStream>>` for example. The latter can still be shut down, it just needs to be done manually.

alexcrichton:
@dwrensha in the meantime, perhaps this could use `Arc<TcpStream>` with a custom implementation of `Read` and `Write`?

alexcrichton:
So with the recent advent of `BiLock`, I think we've got a pretty solid path forward for this:

``` rust
let (r1, w1) = BiLock::new(stream1);
let (r2, w2) = BiLock::new(stream2);

let half1 = copy(r1, w2).and_then(|(_, w, _)| {
    w.lock().and_then(|w| w.shutdown(Shutdown::Write))
});
let half2 = copy(r2, w1).and_then(|(_, w, _)| {
    w.lock().and_then(|w| w.shutdown(Shutdown::Write))
});

let result = half1.join(half2).map(|_| ());

// ...
```

That is, I think we can avoid the TCP-specific methods and instead rely on `BiLock`'s `Deref` impl.

Thoughts?

dwrensha:
@alexcrichton Your suggestion does not immediately work because `BiLock<TcpStream>` does not implement `::std::io::Read` or `::std::io::Write`.

I've posted a new attempt here: #100.

alexcrichton:
Awesome! Closing in favor of https://github.com/tokio-rs/tokio-core/pull/100.

> Your suggestion does not immediately work because BiLock<TcpStream> does not implement ::std::io::Read or ::std::io::Write.

Ah yeah I'm thinking that we should either (a) pub these in futures-rs itself or (b) if we get custom traits in this crate implement them here.

PR #29, opened 2016-09-12
Add `read` free function to read some bytes

3Hren:
Hi!

I've found that there is a way to read from the socket either until EOF or exact number of bytes, but there is no way to read **unknown** number of bytes. This PR fill that gap.

Usage:

``` rust
let rdbuf = [0; 4096].to_vec();
read_some(rd, rdbuf).and_then(|(rd, rdbuf, nread)| {
    println!("read {} bytes: {:?}", nread, &rdbuf[..nread]);    
    Ok(())
})
```

If everything is OK, I'll add documentation.


dwrensha:
Perhaps this should be generalized to `read_at_least()`, which would take a `usize` signifying the minimum number of bytes desired?

3Hren:
I have no strong opinion about your suggestion. Waiting for maintainer's word.

alexcrichton:
Thanks for the PR! We may want to actually stick to the standard library conventions and just call this `read` instead of `read_some`? I forget what the conclusion was about `read_at_least` was in the standard library, but it might be relevant here as well...

src/io/read.rs
```diff
@@ -0,0 +1,65 @@
...
+    type Item = (R, T, usize);
+    type Error = ::std::io::Error;
+
+    fn poll(&mut self) -> Poll<(R, T, usize), ::std::io::Error> {
+        let nread = match self.state {
+            State::Pending { ref mut rd, ref mut buf } => {
+                let buf = buf.as_mut();
+
+                match rd.read(&mut buf[..]) {
+                    Ok(0) => return Err(eof()),
```
alexcrichton: We'd probably want to actually just thread this through? If you don't know how many bytes to get then eof may not always be erroneous.

3Hren: Hmm, if the buffer is zero-sized, then we just want to poll the socket. In this case reading of 0 bytes indicates about either some bytes available or FIN packet received. That's not an error.

On the other side, if the buffer is not zero-sized (which is likely), then reading 0 bytes is surely FIN.

I don't like an idea to make special case for zero-sized buffers. Instead I suggest to make this functionality as an async twin of `Read::read` method.

3Hren:
Just curious, is there any reason to introduce free functions instead of some kind of:

``` rust
trait AsyncRead: Read {
    fn async_read<T>(self, buf: T) -> impl Future<Item=(Self, T, usize)>
        where T: AsMut<[u8]>;

    fn async_read_to_end<T>(self, buf: T) -> impl Future<Item=(Self, T)>
        where T: AsMut<[u8]>;
}
```

?

src/io/read.rs
alexcrichton: Yeah sounds reasonable to me, but that'd be plumbing through `Ok(0)` as `Ok(0)`, right?

alexcrichton:
I'd personally prefer to avoid prematurely adding traits, but we could consider that at a later date yeah.

alexcrichton:
@3Hren what do you think about renaming this to `read`?

3Hren:
Sorry, completely forgot about this PR. Fixed!

src/io/read.rs
```diff
@@ -0,0 +1,62 @@
...
+{
+    type Item = (R, T, usize);
+    type Error = ::std::io::Error;
+
+    fn poll(&mut self) -> Poll<(R, T, usize), ::std::io::Error> {
+        let nread = match self.state {
+            State::Pending { ref mut rd, ref mut buf } => {
+                let buf = buf.as_mut();
+
+                match rd.read(&mut buf[..]) {
```
alexcrichton: I think this can just be `try_nb!`

src/io/read.rs
```diff
@@ -0,0 +1,62 @@
...
+          T: AsMut<[u8]>
+{
+    ReadSome { state: State::Pending { rd: rd, buf: buf } }
+}
+
+/// A future which can be used to easily read available number of bytes to fill
+/// a buffer.
+///
+/// Created by the [`read`] function.
+pub struct ReadSome<R, T> {
```
alexcrichton: s/ReadSome/Read/ here I believe

3Hren:
Fixed again!

Look, is there any way to try to read from the `R` immediately (not after a future is polled), i.e. directly inside the `read` function?

alexcrichton:
Thanks!

arthurprs:
What's the motivation for removing the EOF? Now users will have to check for it everywhere (and possibly hit infinite loops until they spot this, true story)

3Hren:
Just like in `std::io::Read::read` and elsewhere in linux world.

alexcrichton:
Yeah this is the same as the standard library as @3Hren mentioned, which is what this is intended to mirror.

PR #30, opened 2016-09-12
adding sychronous channel implementation

tekjar:
Don't know much about the internals of this crate but this seemed a bit straight forward.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -8,7 +8,7 @@ use std::sync::mpsc::TryRecvError;
 
 use futures::{Poll, Async};
 use futures::stream::Stream;
-use mio::channel;
+use mio::channel::{self, TrySendError};
 
 use reactor::{Handle, PollEvented};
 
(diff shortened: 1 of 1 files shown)


src/channel.rs
```diff
@@ -88,6 +120,59 @@ impl<T> Clone for Sender<T> {
...
+        self.tx.try_send(t).map_err(|e| {
+            match e {
+                channel::TrySendError::Io(e) => e,
+                channel::TrySendError::Disconnected(_) => {
+                    io::Error::new(io::ErrorKind::Other,
+                                   "channel has been disconnected")
+                }
+                channel::TrySendError::Full(_) => {
+                    io::Error::new(io::ErrorKind::Other,
+                                   "channel full")
```
alexcrichton: I think this error case may want to be surfaced in the type itself, it's a standard error case for `try_send` and may be difficult to look at otherwise.

alexcrichton:
Looks good to me! I think we may want to tweak the signature of `try_send` to perhaps:

``` rust
fn try_send(&self, t: T) -> Result<(), TrySendError<T>>;
```

but other than that I think we're good to go.

tekjar:
@alexcrichton  Done. What about the case @carllerche is mentioning where blocking `send` is used inside a `Task`? (Can you add that if necessary, please? I don't know how to proceed from here)

alexcrichton:
Hm, that is indeed an interesting question! That actually brings up more thoughts to mind as well. So ideally `SyncSender::try_send` would then schedule you to receive a notification when it's ready if you're not already. (e.g. that's what you'd expect other I/O objects to do as well). I don't think, though, that we can prevent `send` from being called with a future.

Some of this makes me think that we may not want to wholesale add `sync_channel` just yet. This is essentially just a bounded channel, but in a futures system having a blocking `send` method is pretty dangerous as well...

tekjar:
@alexcrichton How about removing `send` and keeping just `try_send`? Bounded channels are sometimes required and people can handle `BufferFull` error themselves.

carllerche:
One possibility for handling the blocking case would be to check if a task is present and if so, use non-blocking and return an error if a value is not present.

alexcrichton:
Yeah I think removing `send` make sense to me, and only keeping `try_send`. I'm worried, though, that this'd need more support as my first guess would be that if `try_send` returned "would block" my current task would be scheduled to receive a notification when it might not otherwise return "would block".

alexcrichton:
Ok, I'm going to close this for now as we're now trying to consolidate channels into the futures crate rather than the tokio-core crate, but might be a nice addition to have there as well!

PR #35, opened 2016-09-15
Io streams

rrichardson:
OK.  There is a lot going on here, I doubt that the code is in the state where it can be merged, but it makes for decent talking points. 

A couple things:  
**1. Pools and Contiguous memory:**
I don't think we should attempt to guarantee contiguous memory between reads of sockets in the stream.  If someone has that specialized of a need, which seems unlikely, they can implement their own buffer pool to do so.  We could probably add a stream-id parameter to BufferPool::get to allow the pool to try to ensure contiguous pools.  

**2.  Naming.**  
My recommendation is that we change the name of tokio-core::net::TcpStream to TcpSocket.  There isn't really anything special about it that makes it a stream anyways. (I know, probably a long shot for many reasons. )
That aside, I have put the futures::streams implementation into `net::stream::{Tcp, Udp}` 
That seemed like a decent spot for them from a namespace point of view. 

**3. Read to completion.**  
This is something I'm not entirely clear on.  If we don't explicitly say that there isn't any data ready for an io source, will we attempt to read from it at the next poller tick?  I was taking the presence of calls to io.need_read(); to indicate that if that isn't set explicitly, Io::poll_read would continue to indicate that data is waiting.  

 This influences the design pretty heavily because if we have to read to completion for a very fast moving data stream,  there are two issues :  
1. We'll need to be able to return arrays of buffers from stream::poll(). 
2. If we continue to read and produce buffers in a loop, we can starve the other futures. 

I implemented poll for both Tcp and Udp under the optimistic assumption that we can fill only a single buffer at a time and come back to it at the next poll tick.

### DIFF
--- a/src/net/stream_tcp.rs
+++ b/src/net/stream_tcp.rs
@@ -1,0 +1,53 @@
+
+use std::io;
+use std::io::{Read};
+
+use net::TcpStream as NetTcpStream;
+use bytes::MutBuf;
+use bytes::alloc::BufferPool;
+use futures::{Async, Poll};
+use futures::stream::Stream;
+
+
+///
+/// TcpStream
+/// Wraps the UdpSocket and provides a `futures::stream::Stream` implementation
+///
+pub struct TcpStream<S : AsRef<NetTcpStream>, B : BufferPool> {
+    socket: S,
+    pool: B,
+}
+
+impl<S : AsRef<NetTcpStream>, B : BufferPool> TcpStream<S, B> {
+    /// Creates a new TcpStream.  The Buffer pool is a factory of fixed sized
+    /// buffers which is leveraged so that the TcpStream may continually produce
+    /// data.
+    pub fn new(socket: S, b: B) -> TcpStream<S, B> {
+        TcpStream {
+            socket: socket,
+            pool: b
+        }
+    }
+}
+
+impl<S : AsRef<NetTcpStream>, B: BufferPool> Stream for TcpStream<S, B> {
+    type Item = B::Item;
+    type Error = io::Error;
+
+    fn poll(& mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if let Async::NotReady = self.socket.as_ref().poll_read() {
+            return Ok(Async::NotReady)
+        }
+        let mut buf = try!(self.pool.get());
+        match unsafe { self.socket.as_ref().read(buf.mut_bytes()) } {
+            Ok(amt) => { 
+                unsafe { buf.advance(amt) };
+                Ok(Async::Ready(Some(buf))) },
+            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
+                Ok(Async::NotReady)
+            },
+            Err(e) => Err(e)
+        }
+    }
+}
+
--- a/src/net/stream_tcp.rs
+++ b/src/net/stream_tcp.rs
@@ -3,8 +3,7 @@ use std::io;
 use std::io::{Read};
 
 use net::TcpStream as NetTcpStream;
-use bytes::MutBuf;
-use bytes::alloc::BufferPool;
+use bytes::{ MutBuf, BufferPool };
 use futures::{Async, Poll};
 use futures::stream::Stream;
 
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -1,8 +1,9 @@
 use std::io;
 use std::net::{self, SocketAddr, Ipv4Addr, Ipv6Addr};
 use std::fmt;
+use std::mem;
 
-use futures::Async;
+use futures::{Future, Async, Poll};
 use mio;
 
 use reactor::{Handle, PollEvented};
@@ -85,6 +86,34 @@ impl UdpSocket {
             Err(e) => Err(e),
         }
     }
+
+    /// Creates a future that will write the entire contents of the buffer `buf` to
+    /// the stream `a` provided.
+    ///
+    /// The returned future will return after data has been written to the outbound
+    /// socket. 
+    /// The future will resolve to the stream as well as the buffer (for reuse if
+    /// needed).
+    ///
+    /// Any error which happens during writing will cause both the stream and the
+    /// buffer to get destroyed.
+    ///
+    /// The `buf` parameter here only requires the `AsRef<[u8]>` trait, which should
+    /// be broadly applicable to accepting data which can be converted to a slice.
+    /// The `Window` struct is also available in this crate to provide a different
+    /// window into a slice if necessary.
+    pub fn send_dgram<'a, T>(&'a self, buf: T, addr : &'a SocketAddr) -> SendDGram<T>
+        where T: AsRef<[u8]>,
+    {
+        SendDGram {
+            state: UdpState::Writing {
+                sock: self,
+                addr: addr,
+                buf: buf,
+            },
+        }
+    }
+
 
     /// Receives data from the socket. On success, returns the number of bytes
     /// read and the address from whence the data came.
@@ -243,11 +272,68 @@ impl UdpSocket {
     }
 }
 
+
 impl fmt::Debug for UdpSocket {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         self.io.get_ref().fmt(f)
     }
 }
+
+/// A future used to write the entire contents of some data to a stream.
+///
+/// This is created by the [`write_all`] top-level method.
+///
+/// [`write_all`]: fn.write_all.html
+pub struct SendDGram<'a, T> {
+    state: UdpState<'a, T>,
+}
+
+enum UdpState<'a, T> {
+    Writing {
+        sock: &'a UdpSocket,
+        buf: T,
+        addr: &'a SocketAddr,
+    },
+    Empty,
+}
+
+
+fn zero_write() -> io::Error {
+    io::Error::new(io::ErrorKind::WriteZero, "zero-length write")
+}
+
+fn incomplete_write(reason : &str) -> io::Error {
+    io::Error::new(io::ErrorKind::Other, reason)
+}
+
+impl<'a, T> Future for SendDGram<'a, T>
+    where T: AsRef<[u8]>,
+{
+    type Item = T;
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<T, io::Error> {
+        match self.state {
+            UdpState::Writing { ref sock, ref buf, ref addr} => {
+                let buf = buf.as_ref();
+                let n = try_nb!(sock.send_to(&buf, addr));
+                if n == 0 {
+                    return Err(zero_write())
+                }
+                if n != buf.len() {
+                    return Err(incomplete_write("Failed to send entire message in datagram"))
+                }
+            }
+            UdpState::Empty => panic!("poll a SendAllTo after it's done"),
+        }
+
+        match mem::replace(&mut self.state, UdpState::Empty) {
+            UdpState::Writing { buf, .. } => Ok((buf).into()),
+            UdpState::Empty => panic!(),
+        }
+    }
+}
+
 
 #[cfg(unix)]
 mod sys {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,8 @@ log = "0.3"
 mio = "0.6"
 scoped-tls = "0.1.0"
 slab = "0.3"
+bytes = { git = "https://github.com/rrichardson/bytes" }
 
 [dev-dependencies]
 env_logger = "0.3"
+tokio-timer = { git = "https://github.com/tokio-rs/tokio-timer" }
--- a/examples/udp_echo.rs
+++ b/examples/udp_echo.rs
@@ -14,6 +14,7 @@ use tokio_core::net::{ UdpSocket, ByteBufPool };
 use tokio_timer::Timer;
 use std::time::Duration;
 use std::sync::Arc;
+use bytes::Buf;
 
 fn main() {
     let mut core = Core::new().unwrap();
--- a/examples/udp_echo.rs
+++ b/examples/udp_echo.rs
@@ -1,0 +1,52 @@
+
+extern crate futures;
+extern crate bytes;
+extern crate tokio_core;
+extern crate tokio_timer;
+
+use std::net::ToSocketAddrs;
+use std::str;
+use futures::Future;
+use futures::stream::Stream;
+use tokio_core::reactor::Core;
+use tokio_core::net::stream::Udp as UdpStream;
+use tokio_core::net::{ UdpSocket, ByteBufPool }; 
+use tokio_timer::Timer;
+use std::time::Duration;
+use std::sync::Arc;
+
+fn main() {
+    let mut core = Core::new().unwrap();
+    let srvaddr = "127.0.0.1:9999".to_socket_addrs().unwrap().next().unwrap();
+    let cliaddr = "127.0.0.1:9998".to_socket_addrs().unwrap().next().unwrap();
+
+    let clipool = ByteBufPool::new(1024 * 8);
+    let srvpool = ByteBufPool::new(1024 * 8);
+
+    let srv = Arc::new(UdpSocket::bind(&srvaddr, &core.handle()).unwrap());
+    let cli = Arc::new(UdpSocket::bind(&cliaddr, &core.handle()).unwrap());
+   
+    let srv2 = srv.clone();
+    let cli2 = cli.clone();
+
+    let srvstream = UdpStream::new(srv, srvpool);
+    let clistream = UdpStream::new(cli, clipool);
+    
+    let app = cli2.send_dgram(b"PING", &srvaddr).and_then(|_| {
+        let server = srvstream.for_each(|(buf, addr)| { 
+            println!("{}", str::from_utf8(buf.bytes()).unwrap());
+            srv2.send_dgram(b"PONG", &addr).map(|_| ()).wait()
+        });
+        let client = clistream.for_each(|(buf, addr)| { 
+            println!("{}", str::from_utf8(buf.bytes()).unwrap());
+            cli2.send_dgram(b"PING", &addr).map(|_| ()).wait()
+        });
+        server.join(client)
+    });
+
+    let timer = Timer::default(); 
+    let wait = timer.timeout(app, Duration::from_millis(500));
+
+    core.run(wait).unwrap();
+}
+
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,6 +98,7 @@
 extern crate futures;
 extern crate mio;
 extern crate slab;
+extern crate bytes;
 
 #[macro_use]
 extern crate scoped_tls;
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -5,7 +5,52 @@
 
 mod tcp;
 mod udp;
+mod stream_udp;
+mod stream_tcp;
+
+use std::io;
+use std::fmt::{self};
+use bytes::{MutByteBuf};
+use bytes::alloc::BufferPool;
 
 pub use self::tcp::{TcpStream, TcpStreamNew};
 pub use self::tcp::{TcpListener, Incoming};
-pub use self::udp::UdpSocket;
+pub use self::udp::{UdpSocket};
+
+
+/// Implementations of futures::streams for TCP and UDP
+pub mod stream {
+    pub use super::stream_udp::UdpStream as Udp;
+    pub use super::stream_tcp::TcpStream as Tcp;
+}
+
+///
+/// ByteBufPool
+/// Example naieve implementation of the BufferPool
+///
+pub struct ByteBufPool {
+    size : usize
+}
+
+impl ByteBufPool {
+    ///contstruct a new Buffer Pool
+    pub fn new(sz : usize) -> ByteBufPool {
+        ByteBufPool {
+            size : sz
+        }
+    }
+}
+
+impl BufferPool for ByteBufPool {
+    type Item = MutByteBuf;
+    fn get(&self) -> Result<Self::Item, io::Error> {
+        Ok(MutByteBuf::with_capacity(self.size))
+    }
+}
+
+
+impl fmt::Debug for ByteBufPool {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "ByteBufPool {{ size: {} }}", self.size)
+    }
+}
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -8,10 +8,9 @@ mod udp;
 mod stream_udp;
 mod stream_tcp;
 
-use std::io;
 use std::fmt::{self};
-use bytes::{MutByteBuf};
-use bytes::alloc::BufferPool;
+use bytes::buf::{SliceBuf};
+use bytes::{BufferPool, AllocError};
 
 pub use self::tcp::{TcpStream, TcpStreamNew};
 pub use self::tcp::{TcpListener, Incoming};
@@ -42,9 +41,9 @@ impl ByteBufPool {
 }
 
 impl BufferPool for ByteBufPool {
-    type Item = MutByteBuf;
-    fn get(&self) -> Result<Self::Item, io::Error> {
-        Ok(MutByteBuf::with_capacity(self.size))
+    type Item = SliceBuf;
+    fn get(&self) -> Result<Self::Item, AllocError> {
+        Ok(SliceBuf::with_capacity(self.size))
     }
 }
 
--- a/src/net/stream_udp.rs
+++ b/src/net/stream_udp.rs
@@ -3,8 +3,7 @@ use std::io;
 use std::net::SocketAddr;
 use std::convert::AsRef;
 use net::{ UdpSocket };
-use bytes::MutBuf;
-use bytes::alloc::BufferPool;
+use bytes::{ MutBuf, BufferPool };
 use futures::{Async, Poll};
 use futures::stream::Stream;
 
--- a/src/net/stream_udp.rs
+++ b/src/net/stream_udp.rs
@@ -1,0 +1,59 @@
+
+use std::io;
+use std::net::SocketAddr;
+use std::convert::AsRef;
+use net::{ UdpSocket };
+use bytes::MutBuf;
+use bytes::alloc::BufferPool;
+use futures::{Async, Poll};
+use futures::stream::Stream;
+
+///
+/// UdpStream
+/// Wraps the UdpSocket and provides a `futures::stream::Stream` implementation
+///
+pub struct UdpStream<S : AsRef<UdpSocket>, B : BufferPool> {
+    socket: S,
+    pool: B,
+}
+
+impl<S : AsRef<UdpSocket>, B : BufferPool, > UdpStream<S, B> {
+    /// Creates a new UdpStream.  The Buffer pool is a factory of fixed sized
+    /// buffers which is leveraged so that the UdpStream may continually produce
+    /// data.
+    pub fn new(socket: S, b: B) -> UdpStream<S, B> {
+        UdpStream {
+            socket: socket,
+            pool: b,
+        }
+    }
+
+    /// Return a reference to the underlying UDP socket
+    pub fn socket<'a>(&'a self) -> &'a S {
+        &self.socket
+    }
+}
+
+impl<S : AsRef<UdpSocket>, B: BufferPool> Stream for UdpStream<S, B> {
+    type Item = (B::Item, SocketAddr);
+    type Error = io::Error;
+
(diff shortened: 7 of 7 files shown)


rrichardson:
Conversation with @carllerche brought up some interesting points, namely that the AsMut contract, because it is safe, should not present un-initialized bytes.  This means we should explicitly require an unsafe fn mut_buf() -> &mut [u8] in the Buffer trait. While we're at it, we should lift all of https://github.com/carllerche/bytes/blob/master/src/buf/mod.rs#L171-L188 into the definition of Buffer for Tokio. This brings me to my next point...

Library Layering:  It seems, at this point, that tokio-core might want have a Buffer trait and a Buffer pool on which it depends for many operations maybe that's too high level, perhaps futures/streams will want to reference this directly. 

I don't really care where it lives, but I would prefer that BufferPool and Buffer trait remain as traits. I fully intend to plug in my own thread-safe, sequential, recycling allocator for my needs. (e.g. https://github.com/rrichardson/block_alloc)  and other people might want to provide a pool which can make stricter guarantees with regards to contiguous buffers betwixt reads.

rrichardson:
Erg.. there is one other issue which I stupidly did not think about sooner.  The fact that the underlying udp socket is needed for future communication when handling data

e.g. 

``` rust
let stream = Udp(serversocket, pool); 
stream.for_each(|(buf, fromaddr)|  { ...; serversocket(?).send_to(newbuf, fromaddr)); });
```

Duplicating the socket is one option, but not savory.  Passing a ref to the socket along with the data and fromaddr seems tricky from a lifetime perspective. 

Thoughts?

rrichardson:
OK.  One more update.  I added a method and a Futures impl called send_all_to and SendAllTo into the udp socket. These mimick the io::write_all method.  Now an owner of a UdpSocket can send data with a method on the socket. I figured a method was fine, as opposed to a free function, since the functionality is very specific to UDP. 

I also created a try_clone() method which mimicks the mio::udp::try_clone.  I needed 3 copies of the udp client socket to make my test work.  Hints on how to fix that are appreciated.

I created 
https://github.com/rrichardson/tokio-core/blob/io_streams/examples/udp_echo.rs   Which tests the new UDP functionality. 

I am not sure if this is idiomatic futures code, let me know if I should change anything.

tailhook:
> I also created a try_clone() method which mimicks the mio::udp::try_clone. I needed 3 copies of the udp client socket to make my test work. Hints on how to fix that are appreciated.

I believe you can wrap socket into an Arc. As all operations require immutable access to the socket.

---

What's the purpose of the `send_all_to` operation on UDP socket? At least on unixes `sendto` operation accepts the whole UDP message and atomically sends it. And peer receives it as a whole. If you send pieces of the buffer using several `sendto` calls by breaking down message at arbitrary offsets, pieces may be received out of order on the peer and it will not be able to reconstruct the original message.

---

I also wonder if it's useful to split this PR and merge just stream support for UDP first? I.e. it looks like much easier task than getting it right for the TCP.

rrichardson:
@tailhook  -  send_all_to just mimicks the naming convention of the corresponding Tcp function send_all and the Write trait fn write_all(). 

It is purely just a name,  but yea, I agree that the _all_ is a bit superfluous, since send_to will send the entire datagram (if it is under 65k in size)

rrichardson:
speaking of which, I should probably verify that send_all_to will fail if the message len is > 65k.  I think the underlying call will fail, but we should double-check that.

src/net/udp.rs
```diff
@@ -243,12 +283,66 @@ impl UdpSocket {
...
+    where T: AsRef<[u8]>,
+{
+    type Item = T;
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<T, io::Error> {
+        match self.state {
+            UdpState::Writing { ref sock, ref buf, ref addr, ref mut pos } => {
+                let buf = buf.as_ref();
+                while *pos < buf.len() {
```
tailhook: Am I misreading something or is it a loop that may send from arbitrary point in the buffer? Or do you rely on the kernel on never returning anything other than zero or buffer length?

src/net/udp.rs
```diff
@@ -86,6 +98,34 @@ impl UdpSocket {
         }
     }
 
+    /// Creates a future that will write the entire contents of the buffer `buf` to
+    /// the stream `a` provided.
+    ///
+    /// The returned future will not return until all the data has been written, and
```
tailhook: This might be a copy from a TcpStream, but it's misleading nevertheless.

tailhook (commented):
> It is purely just a name

It's not just a name in code and comments. I admit that it might happen to work correctly, but looks wrong anyway.

src/net/udp.rs
rrichardson: You're right.  This will transparently send multiple datagrams.  
We should probably remove the loop and then verify that the the buf doesn't exceed a max datagram size, and probably return an error if it didn't send buf size, as the datagram would be smaller than intended.

alexcrichton:
Thanks for the PR @rrichardson!

Believe it or not this is actually very similar to the early early early versions of futures-based I/O, and I agree that it can be quite useful to think of a TCP stream as simply a stream of incoming buffers (e.g. packets).

Up to this point though we've been wary about having a buffering abstraction in `tokio-core` itself. Even with a generic trait proposed here there's only a small handful of performant implementations (if only one), so we've found that in practice it doesn't end up being very flexible and still binds you to a particular buffering strategy quickly.

Another downside of an approach like this is management of the ownership of the original resource. For example once you create a `TcpStream<B>` it looks like you can't write to it? (similar questions end up plauging other designs as well)

We're still working through the `tokio-proto` layer where you're actually parsing lots of bytes off the wire and dealing with protocol-specific implementation details. That's got lots of questions about buffering and such, but the general trend right now seems to be towards stick with a principle of no buffering and otherwise have adaptors to convert to buffered/future-based interfaces quickly.

I like how this can easily slot in, however, at any layer and give you a stream of buffers immediately. In that sense we may be able to extend this to all `io::Stream` (or `io::Read`) types instead of just `TcpStream`.

I guess my preference would be to perhaps place this in an upper crate for now rather than tokio-core itself, but I'm curious to others' thoughts as well! (cc @aturon and @carllerche)

---

Also as an aside, I agree with @tailhook that we probably don't want to expose `send_to_all`. The semantics of UDP vs TCP are different enough that it seems that any robust/correct application would handle partial writes in a custom fashion.

rrichardson:
Re: Reading from and writing to "streams": When implementing this stuff in the wild, typically those end up being separate objects with separate "loops" which have a queue in the middle. (this is for low-latency situations where you're polling inbound and outbound interfaces)  I think the same abstraction will work here as well, even if they're in the same loop.   I can expand the implementation to a ReadStream and a WriteSink or somesuch.  I have no problem with this going into a higher level crate. 

Re: Buffering.. I think the trait abstraction is tricky, but I'd very much like to solve it.  The problem, as you were probably alluding to, is recycling buffer schemes run into serious lifetime problems.  I beat my head against the wall trying to wedge [this slab allocator](https://github.com/rrichardson/block-alloc-appendbuf/blob/master/src/lib.rs#L286) into the trait that I created for the streams.  The issue is that a buffer's lifetime must be shorter than the pool that created it, which makes it problematic for use in any closure.   I'm not sure if there is a way to unsafely cheat there, while still enforcing the necessary semantics.  I'd probably be okay with a constraint of 'static for the pools.  If someone could lend some advice here I'd appreciate it. 

It sounds like the thing to do then for this PR is reduce it only to the tcp.rs and udp.rs additions (while changing send_all_to to send_to and removing the write loop.  For the streams, would a tokio-streams crate make sense?  I'll look in to generalizing them for io::Read.

carllerche:
Given the lack of clarity re how things will shake up higher up the stack, maybe we should create `tokio-extra`

It could be a staging crate for things that we think could be useful but we don't know if a) they should be part of an existing Tokio crate or a standalone crate and b) unsure of how useful it will be in practice.

alexcrichton:
@aturon, @carllerche, and I talked a bit today and it sounded like `tokio-core` may actually pick up a dependency on `bytes` after all, in which case this may actually fit quite well into the stack (e.g. if we've got it may as well take advantage of it).

I'm not entirely certain what the core combinators for this would want to end up looking like. I think we'd want them a little more abstract than only working on TCP streams and UDP packets. The implementation here though I think is close to what we'd want.

I'm also not 100% sure what to do about `try_clone`. I'd want to investigate what it means to clone a handle already associated with an event loop as I suspect the OS may have weird semantics with where notifications get delivered too and such.

rrichardson:
Glad to hear you're looking at incorporating a dependency on byte/buffers.  My current and next project (for my employer, no less)  both will require a slab style factory for buffers, aiming at 0 runtime allocations.    I have a slab implementation I can apply to the bytes crate. I think the ownership and lifetimes issue will take some finessing, however. 

Re: try_clone and multiple handles :   I know epoll will simply give you an error if you attempt a register on an fd that is already registered.   The more insidious bug there is if the underlying mechanism uses CTL_MOD (re-register), then you can potentially lose interest registrations  

Given the udp_streams's current design, I don't think that there is a way to implement https://github.com/rrichardson/tokio-core/blob/io_streams/examples/udp_echo.rs without try_clone.  However, maybe the correct answer is to pass a write-able thing through to each stream iteration, so that the closure can reply inline without jumping through any hoops.   This might actually solve both the moved handle issue as well as the need for a Sink style stream. 

Write-able things in evented systems typically offer semantics where they buffer up outbound requests until the fd becomes writable.  So this might be a natural fit.

tailhook:
> tokio-core may actually pick up a dependency on bytes after all

While I'm not against `bytes` crate at all. I strongly believe that contiguous buffers should be at least possible if not default. I think this makes parsers simpler and faster.

So is the text above means `FramedIo` will go and `tokio-core` depend on `Framed` and `BlockBuf` ?

rrichardson:
I have to disagree here and say that contiguous buffers are not generally
achievable in practice.  If you need contiguous buffers,  I'd recommend
reading requests into smaller buffers, then copying the payloads into
secondary buffers.   This will likely end up being more efficient than
trying to manage the provision of slabs in sequential order, or allocating
very large chunks.
You would be using the secondary buffers less frequently, and if you pick a
power of 2 as the buffer size, and always use that, then any allocator
should be able to effectively re-use the buffers.

The other option would be to provide a Read-esque interface over a buffer
chain, where each buffer has offsets into the actual payload.
This would mean, of course, that the downstream ecosystem would expect to
operate on the same Read-esque interface, it is a bit too much to ask at
this point, but in the long run this is probably the correct answer.  I say
Read-esque, because it would need to support string and char
searching/scanning style functionality to be performant for parsing and
general purpose use.

As an example,  this isn't the prettiest code, its still a POC, but I
provide a Read interface over a buffer chain, where each buffer has offsets
into its data, I then stream the data through 2 sets of parsers, each of
which expects a read interface.
https://github.com/rrichardson/lout/blob/master/src/gelf.rs#L68,
https://github.com/rrichardson/lout/blob/master/src/gelf.rs#L155

tailhook:
> I have to disagree here and say that contiguous buffers are not generally achievable in practice

Well, I'm not sure why do you think like that. While I could agree that "It's almost impossible to make good contguous buffer for every use case ever", it's certainly a good strategy for many many use cases.

We're successfully doing that in [rotor-http](https://github.com/tailhook/rotor-http/) and almost every other crate on top of `rotor` by using [netbuf](http://tailhook.github.io/netbuf/netbuf/index.html). While rotor is not as battle-tested as nginx or other things, it's of about the same speed as the tests of the tokio-minihttp while having much more full support of HTTP protocol and while having no pooling of buffers (except those inside of jemalloc)

While we don't rely on every http request being fit into contiguous buffers fully, our code that parses HTTP headers and another one that parses chunked encoding does look at slices and can get advantages of SIMD in `httparse` crate because of that buffering strategy.

While contiguous buffers certainly have advantages for users, I don't know how to compare the performance of this sort of thing. I.e. just looping over requests in microbenchmarks isn't too convincing as memory fragmentation, thread contention and other things may add up in the real-world use case.

rrichardson:
I don't see anything wrong with that strategy.  I was confused by referring
to this as a contiguous buffering strategy.   By contiguous I assumed you
meant that multiple allocated buffers would be guaranteed to be provided in
contiguous order, but it sounds like that you mean that every write to 1
allocated buffer is guaranteed to be contiguous (until you reach the
capacity of that buffer). Which for net/parsing applications is a fair
thing to require.  If that is the case, then we are on the same page.

In rotor-http you still have to guess at a max http request size, yes?
There isn't a practical limit on the number of headers, though, so what
happens if you can't fit the entire http request head into your buffer?  Do
you just reject the request?  What is the max size?

My approach is typically to allocate a slab, and then from that allocate
fixed-sized buffers, and these buffers can be large enough to hold your
pathological parsing cases (say 128k for an HTTP Header) because they're
very cheap to use and account for.  The buffers support something akin to
the MutBuf interface, so when you write to them, all writes to the buffer
are contiguous until you reach the max capacity of the buffer. There is no
recourse, however, if you reach the end of the buffer and need more space,
other than copy into a larger buffer and keep going.  This is an expensive
process, so one can make it infrequent by guessing buffer sizes (or
rejecting the pathological cases)

tailhook:
> In rotor-http you still have to guess at a max http request size, yes?

We have a logical limit, not a technical one. For simplicity let's think of a buffer as a `Vec<u8>`. We just allocate `Vec::with_capacity(4096)`. Which works for 99% of requests. If headers don't fit we just grow it. Usually, that means reallocation with data copying, but that's fine enough if we got constants right. And we rely on jemalloc for fast reallocations (i.e. I expect it has a pool of 8k and 16k buffers which can be reused quickly).

> My approach is typically to allocate a slab, and then from that allocate fixed-sized buffers, and these buffers can be large enough to hold your pathological parsing cases (say 128k for an HTTP Header)

This isn't as good as it may appear for the "wild internet" too. Because this means any client opening connection and sending one byte (at least in rotor we don't allocate a buffer until first byte arrives) will allocate 128k of memory. It basically means it's easy to DoS-attack server.

> There is no recourse, however, if you reach the end of the buffer and need more space, other than copy into a larger buffer and keep going.

Sure. But I'd classify most things I'm working with into three categories:
1. When dealing with the **"wild internet"** you should assume that any client may be trying to exhaust your resources, so you keep initial buffers small enough. Growing smoothly. Occasional data copy is not as bad here. Note, it still better than incremental parsing headers here, because we don't allocate lots of small strings which make memory fragmentation.
2. For higher performance and **small messages**, you don't need a very large buffer too. For example, if you have a buffer of 4k and your message size is 64 bytes. You will have 64 messages per buffer where most CPU time would be spent on per-message parsing and processing. So larger buffer will not help too much.
3. For high performance and **large messages**, you need a much better length-prefixed protocol. And the task of the first read is to catch message size. Even in HTTP you will usually get `Content-Length` in first 2048 bytes. Then you can allocate buffer large enough right from the start (but again only if you trust the client, see (1) ) and copying small prefix is not a big performance killer here.

While there are more details here, and probably you can get few percents performance better by more protocol specific enhancements, this kind of handling of large/small messages works well for zeromq for example. (in rotor we assume (1) to be more safe by default, we haven't made any comprehensive studies of how bad it is when there are more large requests than there are small)

rrichardson:
So we both execute a re-allocation strategy,  it is just that I advocated
for a manual one, while yours is transparent.  The good news is that a
buffer pool / buffer trait pair would abstract over these differences
nicely.  It would be easy enough for me to build a multi-slab system which
re-allocates via copying under the hood.

My goal is to fix the max memory footprint for a network service at
start-up, and incorporate buffer availability into the push-back mechanisms.

tailhook:
Yes, that makes sense. My worries were about making the only buffering strategy of a rope of fixed-size chunks.

> My goal is to fix the max memory footprint for a network service at
> start-up, and incorporate buffer availability into the push-back mechanisms.

Sounds intriguing :)

alexcrichton:
@rrichardson perhaps the `try_clone` bits could be left to a later PR? For now you can kinda get around those problems with `Arc<UdpSocket>` (or `Rc`) and a custom implementation of the various methods. I'll also note that `send_to_all` may also want to be left to a later PR (due to the semantics of sending multiple datagrams).

@tailhook note that tokio-core will never _require_ a particular buffering strategy. It's intended that it'll always be as flexible as possible at the lowest layer. What we've found though is that when building upwards in the stack (e.g. a pipelining and/or multiplexing implementation) buffering is required pretty quickly, so we need to pick _something_. 

The `bytes` crate needs an audit for the APIs that it provides as well, but in general I think there's also a need for these sorts of applications in Rust to start coalescing around a buffering strategy (there's a dozen or so crates for this today already on crates.io). The choice of `bytes` is by no mean solidified, it's just the first thing we thought of! More info on this soon as it develops of course.

rrichardson:
@alexcrichton -  I renamed send_all_to to send_dgram and made it send only a single buffer, returning an error if it didn't send the complete buffer in a single dgram. This returns a future. 

<del>I would like to get rid of UdpSocket::try_clone but the way UdpStream is structured, it needs to own the UdpSocket, or at least hold a mutable reference to it for quite a while because it has to poll_read/write it.  I can't think of a way for it to share with anything else.</del>

Should we implement a split() for udp socket which splits the socket into a reader and a writer?  This would solve the immediate problem, I think, it might even generalize well to io Streams of all types. --

I just realized that poll_read and poll_write only require &self.  I'll try to make this work with AsRef<UdpSocket>

rrichardson:
hmm.. rebase and then merge.. not really what I had intended.

rrichardson:
Note that this currently relies on my branch of bytes as it contains the trait for the Buffer Pool.  I have a PR in to bytes but is currently held up.  @carllerche might have some input there. Also due to recent changes it also has a merge conflict.

src/net/stream_tcp.rs
```diff
@@ -0,0 +1,52 @@
...
+    type Item = B::Item;
+    type Error = io::Error;
+
+    fn poll(& mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if let Async::NotReady = self.socket.as_ref().poll_read() {
+            return Ok(Async::NotReady)
+        }
+        let mut buf = try!(self.pool.get());
+        match unsafe { self.socket.as_ref().read(buf.mut_bytes()) } {
+            Ok(amt) => {
```
alexcrichton, line 42: This wants to special case `Ok(0)` and wire it up to `None`, right?

src/net/stream_tcp.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn poll(& mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if let Async::NotReady = self.socket.as_ref().poll_read() {
+            return Ok(Async::NotReady)
+        }
+        let mut buf = try!(self.pool.get());
+        match unsafe { self.socket.as_ref().read(buf.mut_bytes()) } {
+            Ok(amt) => { 
+                unsafe { buf.advance(amt) };
+                Ok(Async::Ready(Some(buf))) },
+            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
```
alexcrichton, line 45: I think the `try_nb!` macro should cover this case?

src/net/udp.rs
```diff
@@ -86,6 +87,34 @@ impl UdpSocket {
...
+    /// needed).
+    ///
+    /// Any error which happens during writing will cause both the stream and the
+    /// buffer to get destroyed.
+    ///
+    /// The `buf` parameter here only requires the `AsRef<[u8]>` trait, which should
+    /// be broadly applicable to accepting data which can be converted to a slice.
+    /// The `Window` struct is also available in this crate to provide a different
+    /// window into a slice if necessary.
+    pub fn send_dgram<'a, T>(&'a self, buf: T, addr : &'a SocketAddr) -> SendDGram<T>
```
alexcrichton, line 105: The name here to me is a little innocous. This is to UDP sockets as `tokio_core::io::write_all` is to TCP streams (modulo the partial write semantics). I wonder if we should name this `std::net::send_to` perhaps to be consistent?

src/net/udp.rs
```diff
@@ -243,12 +272,69 @@ impl UdpSocket {
...
+                let buf = buf.as_ref();
+                let n = try_nb!(sock.send_to(&buf, addr));
+                if n == 0 {
+                    return Err(zero_write())
+                }
+                if n != buf.len() {
+                    return Err(incomplete_write("Failed to send entire message in datagram"))
+                }
+            }
+            UdpState::Empty => panic!("poll a SendAllTo after it's done"),
```
alexcrichton, line 327: s/SendAllTo/SendDGram/

alexcrichton (commented):
Sorry for the delay @rrichardson! I think we'll want to be sure to sort out the `bytes` story before merging this, but it's looking otherwise ready to go to me!

src/net/udp.rs
rrichardson: For ergonomics, I kind of prefer it as a method of the UdpSocket struct as opposed to a free function, but as a method it conflicted with the existing send_to method which not a Stream or Future.   

However, this does pose problems for API layering.  I can move this out to be a public function in tokio::core::io, implemented in the stream_udp.rs file.

src/net/stream_tcp.rs
rrichardson: It definitely does, but not if we want to handle the special case of Ok(0)
Is it part of the contract that we deliver Ok(Ready(None)) on a 0 read, or will the upstream consumer figure out that there is a problem?

src/net/stream_tcp.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use net::TcpStream as NetTcpStream;
+use bytes::{ MutBuf, BufferPool };
+use futures::{Async, Poll};
+use futures::stream::Stream;
+
+
+///
+/// TcpStream
+/// Wraps the UdpSocket and provides a `futures::stream::Stream` implementation
```
Ralith, line 13: s/`UdpSocket`/`net::TcpStream`/ or similar

alexcrichton:
Ok, coming back to this after recent developments.

I think the `TcpStream` improvements are basically equivalent to [`Io::framed`](https://tokio-rs.github.io/tokio-core/tokio_core/io/trait.Io.html#method.framed) now, right? That is, most of the time you'll want to end up parsing data out anyway, so this is providing the ability to do so (and providing a buffer for you). Additionally, it also provides the ability to write back into the `TcpStream` with a `Sink` implementation that `Encode` takes.

For UDP, however, I think this is still worthwhile. I wonder if perhaps we could add a `framed` implementation for UDP that's UDP specific? Perhaps something like:

``` rust
impl UdpSocket {
    pub fn frame<D, E>(self) -> FramedUdp<D, E> 
        where D: UdpDecode,
              E: UdpEncode,
    { 
        // ...
    }
}

pub trait UdpDecode: Sized {
    fn decode(src: &SocketAddr, buf: &mut Vec<u8>) -> io::Result<Self>;
}

pub trait UdpEncode: Sized {
    fn encode(self, dst: &mut Vec<u8>) -> SocketAddr;
}

impl<D, E> Stream for FramedUdp<D, E> {
    type Item = D;
    type Error = io::Error;

    // ...
}

impl<D, E> Sink for FramedUdp<D, E> {
    type SinkItem = E;
    type SinkError = io::Error;

    // ...
}
```

I don't think we can reuse the standard encode/decode traits that are used on streams due to the datagram nature of UDP, but in theory these traits should essentially embody the same paradigm?

@rrichardson thoughts on a strategy like that?

rrichardson:
I need to go back and look at Framed in detail, I've not really paid much attention to it yet, so pardon my ignorance, but why does this return a SocketAddr?

 fn encode(self, dst: &mut Vec<u8>) -> SocketAddr;

My concern at this point, is how to pass through the addresses for both send_to and recv_from? 

<del>I think the most straightforward approach is to make the address a parameter in the Encode and Decode traits.</del>  It might also be possible to make Stream and Sink take a tuple of (SocketAddr, D/E)

Also, berkeley sockets UDP does support the notion of "connection" so that the outbound address can be fixed and 'send' is called.  We might want to support that as well.

alexcrichton:
Oh right yeah so the encode half would be for every message received through the sink it'd know where to send that message. That is, the returned `SocketAddr` is where to send the bytes that were just encoded into the `&mut Vec<u8>`.

> My concern at this point, is how to pass through the addresses for both send_to and recv_from? 

Right yeah, this is definitely important. That's why in `Decode` (which is called after a `recv_from`) it's passed the address. That address would be where the data in `&mut Vec<u8>` came from. The address to pass to `send_to` is then returned by `Encode`.

> Also, berkeley sockets UDP does support the notion of "connection" so that the outbound address can be fixed and 'send' is called. We might want to support that as well. 

Perhaps yeah, but I figure we can start out with being explicit and perhaps branch into that area later?

rrichardson:
The latest interface looks good to me, with a couple reservations:   

Would it be EasyBuf instead of vec<u8> though?  EasyBuf definitely solves a significant challenge in my log/event router.   

Also I assume that the lack of Option<Self> in the decode return implies that a datagram will always contain a complete framed message, which isn't always the case. 

So maybe decode should look like : 

``` rust
pub trait UdpDecode: Sized {
    fn decode(src: &SocketAddr, buf: EasyBuf) -> io::Result<Option<Self>>, Error>;
}
```

I can take a whack at implementing this, although I probably won't be able to get to it until Tuesday.

alexcrichton:
Oh if there's a use case for `EasyBuf` that sounds good to me. That was primarily intended for TCP where you may receive multiple requests in one read, whereas with UDP it's always packet-based, right?

Similarly the `Option` was for reading partial responses with TCP but with UDP you have to receive the whole thing in one go, right?

rrichardson:
Because of fragmentation, most UDP protocols still keep their message size to < 1500 bytes (if they're intended to go over the interwebs).   Because of this, chunking is a very common idiom for UDP protocols. For example: Gelf, the Graylog protocol, streams compressed then chunked data over UDP.

1 Question : 

For Lout, I'd like decode to emit a referenced counted buffer, because the received events will likely go to multiple downstream output plugins.  It looks like EasyBuf doesn't support a top-level clone() function, but it should be pretty easy to add, yes? 

Also, I still harbor notions of utilizing custom allocator pools some time in the future.  This means that EasyBuf, or something like it, will need to be able to contain a pointer back to its 'allocator'.   What are your thoughts on that?

tailhook:
> Because of fragmentation, most UDP protocols still keep their message size to < 1500 bytes (if they're intended to go over the interwebs). Because of this, chunking is a very common idiom for UDP protocols.

It's not same as for TCP anyway. Because for TCP you can read more data in the same buffer. For UDP you must read packet first and then you can find out which segment you just read and concatenate it to one of the segment read earlier.

And because there are `mut Vec` passed everywhere you can fetch those vecs from pool.

The only issue with the example above is that you can't return object which refers to slices of the buffer (using `EasyBuf::drain_to`). Not sure how common is will be.

rrichardson:
Thanks, @tailhook , Again, I don't fully understand framed yet, but does Option<Self>  indicate that a Message is complete, or that there is more data available after this current message? 

If it is just plain mut Vecs, it will be tricky to orchistrate fetching them from a pool, because they will need to maintain a reference to the pool whence they came.  It makes lifetimes tricky, but I don't think there is a way around it.

tailhook:
>  Again, I don't fully understand framed yet, but does Option indicate that a Message is complete, or that there is more data available after this current message?

None -- means that current buffer contains no full frames. In other words, decoder asks to try again when more bytes are are received.

When decoder returns `Some(x)` -- it means there is a frame. And `Framed` implementation calls `decode` many times, expecting that more frames might be decoded, until `None` is returned.

alexcrichton:
@rrichardson yeah the `Encode` and `Decode` traits aren't intended to be the end-all-be-all of how protocols are parsed and serialized. They're rather intended as an easy step up to get something working quickly, and then the types can be implemented in a more specialized manner if the perf becomes a problem.

We'd like to use the `bytes` types here eventually but for now (at the 0.1 stage) we're thinking of just sticking with `Vec<u8>` or `EasyBuf` which allows many applications to work efficiently, but by no means all. 

In general a `Framed` just gives you a `Sink` and a `Stream` later on, so the whole purpose of `Decode` and `Encode` is to provide as little logic as possible to get to a sink/stream. If you get to a sink/stream in a different fashion, however (e.g. via a custom impl) that'd also work!

rrichardson:
OK.  Everything is pretty much implemented, but I'm unsure about the Bilock for TCP and UDP Stream/Sink.  It seems to me that in most cases, sending and receiving simultaneously on most all Fds is perfectly OK.  If its not, the OS typically brokers the connection, failing that, we can use the Bilock on specific implementations where we know it is necessary. 

The current implementation could lead to slowdowns, where we're needlessly limiting parallelism.  In addition, it seems like its the wrong abstraction, when what we need is to ensure that multiple copies of Stream or Sink are not used simultaneously.  

I'm tempted to just clone the socket and leave out the Bilock for FramedUdp.  We could even create separate Arcs for each of the Stream and Sink. 

Thoughts?

rrichardson:
closing in favor of https://github.com/tokio-rs/tokio-core/pull/109

copying latest discussion there

PR #36, opened 2016-09-19
Little correction of the example

mboecker:
Instead of calling the method .handle() twice, just re-use the value from above.


carllerche:
Thanks! Merged here: f41c55f9f7b9b4b74b35e9db2516a70b05c21d2f

Issue #40, opened 2016-09-22
Add Handle::spawn_fn


oberien:
As discussed with @carllerche on gitter, `Handle::spawn_fn` would be nice to have. It allows two futures working with `ReadHalf` and `WriteHalf` to be run on the same task without wrapping `handle.spawn(futures::lazy(…))` around.

alexcrichton:
Sure!

PR #41, opened 2016-09-23
Add Chat example

oberien:


alexcrichton:
Nice idea! 

I wonder if perhaps this could be a line-based chat server? That way you could easily play around with it on the command line perhaps.

oberien:
Wanted to that first, but therefore `tokio_core::read_until` is missing. To keep the example simple I decided to not implement it. Will open an issue on it. I also need to adapt this example to use #40.

alexcrichton:
Perhaps this PR could add `read_until` as well? Seems reasonable to reflect libstd in that respect to me at least.

oberien:
I created a separate PR for `read_until` in #48. Could you please accept that one before this one?
I also adapted the example to `Handle::spawn_fn`, so if everything seems fine to you, I think this PR is finished.

oberien:
I rebased this branch to use the latest commit from #48. I think this one can now be merged.

alexcrichton:
Thanks @oberien! Looking a little more into this, I think there's still one remaining bug with `read_until` usage here where it may read multiple lines at once. Perhaps, like `std::io`, we could add a `lines` combinator which returns a `Stream<Item=String, Error=io::Error>`? That could then handle the logic of perhaps reading more than one line at a time.

Also, perhaps as part of the chat example the messages received could be proxied to all connections _except_ the one that sent it?

oberien:
Even though `lines` is an amazing idea, which I'll add in a separate PR, as far as i can understand `BufRead::read_until` [returns everything up to the **first** occurrence of the delimiter](https://doc.rust-lang.org/src/std/up/src/libstd/io/mod.rs.html#1137). In that code `memchr::memchr` is used, which ["Returns the index corresponding to the first occurrence of `needle` in `haystack`, or `None` if one is not found"](http://burntsushi.net/rustdoc/memchr/fn.memchr.html). Therefore `read_until('\n' as u8)` should only return up to exactly one line.

Preventing the sender from being sent to would be very easy to implement. But for example XMPP also sends back messages to the sender, so it'll only be displayed, if the server received the message. That way the client knows if a message is sent correctly.

carllerche:
It is sounding a bit like there is duplication around [tokio-line](github.com/tokio-rs/tokio-line). It would probably be worth figuring out how to organize the line based examples.

alexcrichton:
@oberien oh right, good point about `read_until`! I totally forgot those semantics...

I also figure that because you'd probably test out this sort of example with tools like `telnet` or `nc` that it may be beneficial to not send it back to the sender as you'd just see what you just typed on the terminal already.

@carllerche I think it's good to have examples at multiple layers of the stack, however. Examples from tokio-line would likely be more full-featured, whereas this is just a simple "how to spawn and read" tasks in a reactor.

oberien:
Changed it to the suggested behaviour of sending to everyone except the sender.

This example is showing how to write something to a connection, which is handled by a different task.

examples/chat.rs
```diff
@@ -0,0 +1,96 @@
...
+    let addr = addr.parse().unwrap();
+    // We are single-threaded, so we can just use Rc and RefCell.
+    let connections = Rc::new(RefCell::new(HashMap::new()));
+
+    let mut core = Core::new().unwrap();
+    let handle = core.handle();
+    let socket = TcpListener::bind(&addr, &handle).unwrap();
+    println!("Listening on: {}", addr);
+
+    let connections = connections.clone();
```
alexcrichton: I think this clone isn't necessary as it's not referenced outside the closure

examples/chat.rs
```diff
@@ -0,0 +1,96 @@
...
+    let future = socket.incoming().for_each(move |(stream, addr)| {
+        let connections = connections.clone();
+        let handle_inner = handle.clone();
```
alexcrichton, line 33: It's ok to defer these clones until they're needed as opposed to happening immediately.

examples/chat.rs
```diff
@@ -0,0 +1,96 @@
...
+        // We create a new future in which we create all other futures.
+        // This makes `stream` be bound on the `lazy` future's task, allowing
+        // `ReadHalf` and `WriteHalf` to be shared between inner futures.
+        handle.spawn_fn(move || {
```
alexcrichton: One part of futures is helping with "rightward drift", but this file seems to unfortunately suffer from that. Could this perhaps be refactored to not be so indented?

examples/chat.rs
```diff
@@ -0,0 +1,96 @@
...
+                        // For each open connection except the sender, send the string via the channel
+                        for tx in connections.borrow_mut().iter().filter(|&(&k,_)| k != addr).map(|(_,v)| v) {
+                            tx.send(message.clone()).unwrap();
+                        }
+                        futures::finished((Some(reader),None))
+                    })
+            });
+
+            // Whenever we receive a string on the Receiver, we write it to `WriteHalf<TcpStream>`.
+            let socket_writer = rx.fold((None, Some(writer)), move |(_, writer), msg| {
```
alexcrichton: This is pretty gnarly, perhaps it could be written as:

``` rust
let writer = rx.fold(writer, |wr, msg| {
    io::write_all(wr, msg)
});
```

and then transforms can happen later if need be.

examples/chat.rs
```diff
@@ -0,0 +1,96 @@
...
+            connections.borrow_mut().insert(addr, tx);
+
+            let connections_inner = connections.clone();
+            // https://users.rust-lang.org/t/loop-futures-for-client-handling/6950/2
+            // We have an endless loop reading from a client.
+            // In order to fuse the reading and writing futures in the end, we need to have the same
+            // output type. Therefore we use `(Option<BufReader<ReadHalf<TcpStream>>>,
+            // Option<WriteHalf<TcpStream>>)`.
+            let reader = BufReader::new(reader);
+            let socket_reader = stream::iter::<_, _, std::io::Error>(iter::repeat(()).map(Ok)).fold((Some(reader),None), move |(reader, _), _| {
```
alexcrichton: This is quite verbose! Perhaps we can reduce this a bit? 

There's a few things in play here:
- I think the return value of this future should be normalized later, instead this should in general just be a `fold` operation.
- Could the "infinite iterator" be defined on its own line? 

Note that this is why I was thinking it'd be good to add `io::lines` as it'd be a perfect fit for this use case.

examples/chat.rs
```diff
@@ -0,0 +1,96 @@
...
+                let reader = reader.unwrap();
+                let connections = connections_inner.clone();
+                // read and parse length prefix
+                io::read_until(reader, '\n' as u8, vec![])
+                    .and_then(|(reader, vec)| futures::lazy(|| {
```
alexcrichton: I don't think the lazy bit here is necessary?

examples/chat.rs
```diff
@@ -0,0 +1,96 @@
...
+                        // EOF was hit without reading a delimiter
+                        if vec.len() == 0 {
```
alexcrichton: Shouldn't EOF gracefully terminate the connection instead of killing everything?

examples/chat.rs
```diff
@@ -0,0 +1,96 @@
...
+                            futures::failed((std::io::Error::new(std::io::ErrorKind::BrokenPipe, "Broken Pipe"))).boxed()
+                        } else {
+                            futures::finished((reader, vec)).boxed()
+                        }
+                    }))
+                    // convert bytes into string
+                    .map(|(reader, vec)| (reader, String::from_utf8(vec).unwrap()))
```
alexcrichton: Perhaps this could be handled gracefully by sending a message back indicating "you didn't send utf-8"?

examples/chat.rs
oberien: A TcpStream is gracefully terminated by dropping it. As `Io::split` on `TcpStream` creates a `Rc<TcpStream>` and a clone of it, I need to drop `ReadHalf<TcpStream>` and `WriteHalf<TcpStream>` for the TcpStream to be dropped. I don't have access to the `WriteHalf` when hitting EOF. To drop it, I need to finish the `Task` of these futures. The easiest way to do so is to return an error, which will cause `then` to be executed and the task to be finished.

Is there a better way to do this, which I'm missing?

examples/chat.rs
oberien: I'm using moving closures. This means that I need to clone everything in the outer moving closure before creating the inner one. Therefore I don't see the possibility to defer these clones.

examples/chat.rs
oberien: I Removed it.

examples/chat.rs
oberien: I tried improving the rightward drift to comply with the conventions of ≤ 99 characters. I don't see a better way to reduce the number of closures as they need to be moving ones. therefore I can't define them elsewhere to reduce indentation. Is there a better way than defining moving closures inside of moving closures in this example?

examples/chat.rs
oberien: I used multiple lines for this as suggested to make it more readable. Even though I like the idea of `io::lines` and would like to add such a convenience wrapper, I have the following two arguments against it:
- Examples in tokio-core should show how to create an infinite loop.
- Adding such a method to tokio-core could interfere with tokio-line (@carllerche).

As I think it may need more discussion I created #53.

examples/chat.rs
oberien: Removed.

examples/chat.rs
oberien: Nice idea, implemented in the latest commit.

examples/chat.rs
oberien: This definitely makes the code clearer and more readable. Done in the latest commit.

examples/chat.rs
```diff
@@ -0,0 +1,111 @@
...
+            // In order to fuse the reading and writing futures in the end, we need to have the
+            // same output type. Therefore we use `(Option<BufReader<ReadHalf<TcpStream>>>,
+            // Option<WriteHalf<TcpStream>>)`.
+            let socket_reader = socket_reader.map(|reader| (Some(reader), None));
+            let socket_writer = socket_writer.map(|writer| (None, Some(writer)));
+
+            let amt = socket_reader.select(socket_writer);
+            amt.then(move |_| {
+                connections.borrow_mut().remove(&addr);
+                println!("Connection {:?} closed.", addr);
```
alexcrichton: s/{:?}/{}/

examples/chat.rs
```diff
@@ -0,0 +1,111 @@
...
+                let amt = amt.map(|(reader, vec)| (reader, String::from_utf8(vec)));
+                amt.and_then(move |(reader, message)| {
+                    println!("{}: {:?}", addr, message);
+                    let conns = connections.borrow_mut();
+                    if let Ok(msg) = message {
+                        // For each open connection except the sender, send the string
+                        // via the channel
+                        let iter = conns.iter().filter(|&(&k,_)| k != addr).map(|(_,v)| v);
+                        for tx in iter {
+                            tx.send(msg.clone()).unwrap();
```
alexcrichton: perhaps this could also send back the address each message comes from?

examples/chat.rs
```diff
@@ -0,0 +1,111 @@
...
+                let amt = io::write_all(writer, msg.into_bytes());
+                let amt = amt.map(|(writer, _)| writer);
+                amt
+            });
+
+            // In order to fuse the reading and writing futures in the end, we need to have the
+            // same output type. Therefore we use `(Option<BufReader<ReadHalf<TcpStream>>>,
+            // Option<WriteHalf<TcpStream>>)`.
+            let socket_reader = socket_reader.map(|reader| (Some(reader), None));
+            let socket_writer = socket_writer.map(|writer| (None, Some(writer)));
```
alexcrichton: These values aren't actually needed, so perhaps they could get mapped to `()`?

examples/chat.rs
```diff
@@ -0,0 +1,111 @@
...
+                    if vec.len() == 0 {
+                        let err = Error::new(ErrorKind::BrokenPipe, "Broken Pipe");
+                        futures::failed(err).boxed()
+                    } else {
+                        futures::finished((reader, vec)).boxed()
+                    }
+                });
+                // convert bytes into string
+                let amt = amt.map(|(reader, vec)| (reader, String::from_utf8(vec)));
+                amt.and_then(move |(reader, message)| {
```
alexcrichton: Perhaps this could just be `map` that maps to a `()` value as this doesn't need to block?

examples/chat.rs
```diff
@@ -0,0 +1,111 @@
...
+                // read line
+                let amt = io::read_until(reader, '\n' as u8, vec![]);
+                // check if we hit EOF and need to close the connection
+                let amt = amt.and_then(|(reader, vec)| {
+                    // EOF was hit without reading a delimiter
+                    if vec.len() == 0 {
+                        let err = Error::new(ErrorKind::BrokenPipe, "Broken Pipe");
+                        futures::failed(err).boxed()
+                    } else {
+                        futures::finished((reader, vec)).boxed()
```
alexcrichton: Instead of `finished`/`failed` here could `Result` be used?

examples/chat.rs
```diff
@@ -0,0 +1,111 @@
...
+    let socket = TcpListener::bind(&addr, &handle).unwrap();
+    println!("Listening on: {}", addr);
+
+    let future = socket.incoming().for_each(move |(stream, addr)| {
+        let connections = connections.clone();
+        let handle_inner = handle.clone();
+        // We create a new future in which we create all other futures.
+        // This makes `stream` be bound on the outer future's task, allowing
+        // `ReadHalf` and `WriteHalf` to be shared between inner futures.
+        handle.spawn_fn(move || {
```
alexcrichton: Perhaps this could get pushed down to the very bottom? It'd be nice to not indent everything inside of a `spawn_fn` if possible

examples/chat.rs
oberien: As described [earlier](https://github.com/tokio-rs/tokio-core/pull/41#discussion_r82183033) I need the future to error into `then` to close the connection. Returning `Err` would result in `Async::Ready<Err>`, which does not make the future fail and call `select` and `then`.

examples/chat.rs
oberien: I moved `spawn_fn` to the bottom. Unfortunately this does not change the indentation, which stems from the closure. I need this closure so that when spawing it, `Io::split`, `socket_reader` and `socket_writer` are executed in the same task.

examples/chat.rs
oberien: Done

examples/chat.rs
oberien: Done

examples/chat.rs
oberien: That makes the code way more readable - Done.

examples/chat.rs
oberien: As it does not need blocking, I changed it to `map`. As it is called inside `fold` it still returns the reader. - Done

examples/chat.rs
alexcrichton: Oh here I just mean that you can return `Err(err)` and `Ok((reader, vec))` as `Result` implements `IntoFuture`

examples/chat.rs
```diff
@@ -0,0 +1,112 @@
...
+            let socket_writer = socket_writer.map(|_| ());
+
+            let amt = socket_reader.select(socket_writer);
+            amt.then(move |_| {
+                connections.borrow_mut().remove(&addr);
+                println!("Connection {} closed.", addr);
+                Ok(())
+            })
+        };
+        handle.spawn_fn(main_fn);
```
alexcrichton, line 105: Hm this doesn't really solve the problem of "let's see if we can un-indent things", so creating a variable just for `main_fn` likely isn't worth it. Perhaps a list of futures can be created though?

examples/chat.rs
```diff
@@ -0,0 +1,112 @@
...
+            connections.borrow_mut().insert(addr, tx);
+
+            let reader = BufReader::new(reader);
+            let connections_inner = connections.clone();
+            // https://users.rust-lang.org/t/loop-futures-for-client-handling/6950/2
+            // First we need to get an infinite iterator
+            let iter = stream::iter::<_, _, std::io::Error>(iter::repeat(()).map(Ok));
+            // Then we fold it as infinite loop
+            let socket_reader = iter.fold(reader, move |reader, _| {
+                let connections = connections_inner.clone();
```
alexcrichton: I think this clone is needed, right?

oberien: Even though it makes no sense to me, it is needed. If I leave it out and instead use `connections_inner`, I get `cannot move out of captured outer variable in an`FnMut`closure`.

examples/chat.rs
oberien: Could you please explain what you mean with a list of futures? An idea I have to reduce rightward drift is to wrap `connections` in a struct implementing all relevant functions. But this would make the code harder to understand. I'd like to keep the examples and their circular complexity as low as possible to make them understandable at first glance, which would not be the case with my idea. At the same time examples should show good code quality, which unfortunately is not the case with the code right now because of rightward drift.

examples/chat.rs
oberien: Didn't know about `Result` implementing `IntoFuture`. Thank you. Done

alexcrichton:
Thanks again for the example!

spinda:
Since this uses `BufReader`, won't the chat server have to wait until enough messages have been received to fill the buffer before forwarding any of them?

oberien:
[`BufReader::read_until`](https://doc.rust-lang.org/src/std/up/src/libstd/io/mod.rs.html#1313) uses [an internal `read_until`](https://doc.rust-lang.org/src/std/up/src/libstd/io/mod.rs.html#1125) function. That function calls [`fill_buf`](https://doc.rust-lang.org/src/std/up/src/libstd/io/buffered.rs.html#185), which uses the wrapped reader's `read` function for BufReader. The inner reader is mio's `TcpStream`, which returns the number of bytes read and does not block. Therefore `Ok(n)` is returned and the `read_until` method checks if the delimiter was hit. No blocking should be needed or happen.

Issue #37, opened 2016-09-20
Support recovery from accept errors in TcpListener::incoming()


ghost:
Currently whenever accept returns error the incoming stream will finish in a
failed state. It is not possible to continue accepting further connections
without creating a new TcpListener and rebinding it.

This is highly problematic especially that accept errors are not uncommon, and
easily produced by malicious client.

One solution, which seems to me quite ergonomic, would be to change the item
type of Incoming to `Result<(TcpStream, SocketAddr), io::Error>`. Stream
consumers would be free to handle errors exactly as they wish.

The only serious limitation of this interface is the fact that you can't get
TcpListener back. Such requirement seems uncommon, and supporting it has some
non-trivial ergonomic cost. This limitation is of course pre-existing.

ghost:
Additional notes, after experimenting with approach described above. Using
result item type, the previous behaviour can be simulated with following code:

``` rust
listener.incoming().and_then(|r| r)
```

Ignoring all accept errors is equally trivial, and can be done with:

``` rust
listener.incoming().map_filter(|r| r.ok())
```

I have implementation for this proposal at following branch:
https://github.com/tmiasko/tokio-core/tree/tcp-listener-incoming-result

alexcrichton:
Stream errors are currently recoverable in the sense that they don't terminate the stream, Unlike a future, a stream only terminates with `Async::Ready(None)`. 

That is, you can use `or_else` to deal with errors (or any other combinator for this). Does that make sense?

ghost:
Yeah, thanks!

lilydjwg:
@tmiasko How did you recover with `or_else`?

I tried but it requires me to produce a `(TcpStream, SocketAddr)`. I tried to `map` the `Ok` part to `Option<(TcpStream, SocketAddr)>` so I can represent errors with `None`,  but then `.for_each` disappeared, because it's not something can be turned into a Future.

ghost:
@lilydjwg, using `or_else` immediately after `incoming()` is indeed problematic because the types have to agree, but you can `map` it first into a stream of `()` and then recover from errors with `or_else`. Alternatively you could use `then` and process both success and error cases together.

PR #42, opened 2016-09-23
Return a stream of accept results from TcpListener::incoming().

ghost:
Addresses issue #37.


alexcrichton:
I've [left a comment](https://github.com/tokio-rs/tokio-core/issues/37#issuecomment-249447194) on #37, perhaps we can continue discussion there?

PR #48, opened 2016-09-29
feat(io): Add `read_until`

oberien:


src/io/read_until.rs
```diff
@@ -0,0 +1,74 @@
...
+        match self.state {
+            State::Reading { ref mut a, byte, ref mut buf } => {
+                // If we get `Ok(n)`, then we know the stream hit EOF or the delimiter.
+                // If we got EOF (n == 0), then we return an error.
+                // If we hit "would block" then all the read data so far
+                // is in our buffer, and otherwise we propagate errors.
+                // If we reached our delimiter (n != 0), then we are finished.
+                let n = try_nb!(a.read_until(byte, buf));
+                if n == 0 {
+                    return Err(io::Error::new(io::ErrorKind::BrokenPipe, "Broken Pipe"));
```
alexcrichton: I think this isn't an error if we've read any bytes so far, in which case we can just return the buffer.

oberien: I added a check `if n == 0 && buf.len() == 0` to only catch the case where the reader hit EOF.

alexcrichton: ah actually I think it needs to save off the size of `buf.len()` before the I/O even starts, because the vector could have data in it before we call `read_until`. (then the length after n ==0 is checked against the original length)

ghost:
The `BufRead::read_until` reads until either delimiter or EOF. In both cases it
would return `Ok(n)` with number of bytes read. If goal is to make an
equivalent to one from std library then I would suggest updating the change
accordingly. 

On the other hand, if succeeding only when the delimiter is found is
intentional, I would refrain from calling this an async equivalent of
`BufRead::read_until` without making clear what differs between the two. And in
the EOF case, returning `UnexpectedEof` error kind seems more appropriate.

oberien:
My main idea is to make the async version of `read_until` the same as `BufRead::read_until`. The problem I had when testing with the [chat example](41) was that if a `TcpStream` was closed by the client, `read_until` did not return an error, but instead an empty buffer.
From my point of view, if nothing can be read (e.g. the connection is closed), an error should be raised instead of returning an empty buffer. Therefore I added the check to see if nothing has been read. This is done by checking if `n == 0` to verify, that in the last iteration of `ReadUntil::poll` nothing new has been read, and with `buf.len() == 0` to verify that also in previous iteration nothing has been read.

Let's make an example, where read_until is called with "\n" as delimiter:

| first iteration | check | second iteration | check | third iteration | check | result |
| --- | --- | --- | --- | --- | --- | --- |
| `"foo\n"` | `n == 4 len == 4` |  |  |  |  | `"foo\n"` |
| `"foo"` | `n == 3 len == 3` | `"bar\n"` | `n == 4 len == 7` |  |  | `"foobar\n"` |
| `"foo"` | `n == 3 len == 3` | `""` | `n == 0 len == 3` | `"bar\n"` | `n == 4 len == 7` | `"foobar\n"` |
| `""` | `n == 0 len == 0` |  |  |  |  | `Err(broken pipe)` |

Actually for each case only the last check would be done, as `WouldBlock` makes the function return early from `try_nb` without the check being done.

Other than this change, the functionality should equal `BufRead::read_until`.

alexcrichton:
Oh so I'm under the impression that we want to mirror `BufRead:read_until` as much as possible, so in that sense I don't think we should manufacture an error here?

oberien:
Today in today I learned: `BufRead::read_line` returns `Ok(0)` and `""` if the wrapped stream is dead just like any other language… (I literally just realized that o.O) I'll change `tokio_core::io::read_until` to match `BufRead::read_until`.

oberien:
Done. To keep the discussion useful, I added a new commit. While merging it would be best to squash them :)

arthurprs:
This is a good addition. Is this the right way to do it though?

Maybe it should drop the BufRead requirement and as a side-effect possibly read stuff after the terminator (it'll have to return the index where the terminator is at too).

oberien:
That's exactly the problem. We have multiple options here:
1. We already take a `BufRead` and return it, which is how I implemented it currently. This has the disadvantage of the need to wrap `Read` with `BufReader::new`.
2. We take a `Read` and return `BufReader<Read>`, which has multiple disadvantages: 
   - You cannot use the method in a loop, because every call changes the type.
   - Calling the method multiple times results in `BufReader<BufReader<BufReader<Read>>>`.
3. We take `Read`, wrap it in `BufReader<Read>` and convert it back to `Read` before returning. This means that either we lose a part of the buffer or the buffer we return contains more bytes than up to the delimiter. This means that `read_until` is not an async version of `BufRead::read_until`.

Given we use the 3rd one, if we want to implement `read_line` based on `read_until`, we would have the problem that we read beyond `\n`, which is not what you would expect from `read_line`.

arthurprs:
It's a known tradeoff, possibly reading more than necessary is what rotor and boost::asio do too. Trying to mimic/forcing a BufRead is probably not a good idea at this level of abstraction.

alexcrichton:
Thanks @oberien!

@arthurprs this likely isn't going to be the foundation of most async I/O reads/writes. Instead it's intended to be a convenient adapter, if necessary, available to you (as it's basically got one canonical implementation). That's the thinking along the lines of the other functions like `write_all` and `read_to_end`, etc.

PR #52, opened 2016-10-05
Implement `tokio_core::reactor::Interval`

tailhook:
An easy and I think the idiomatic way to handle recurring timers.

Also, docstring of Timeout is outdated, right?

> Timeouts are created through the LoopHandle::timeout or LoopHandle::timeout_at methods indicating when a timeout should fire at

### DIFF
--- a/src/reactor/interval.rs
+++ b/src/reactor/interval.rs
@@ -1,0 +1,171 @@
+//! Support for creating futures that represent intervals.
+//!
+//! This module contains the `Interval` type which is a stream that will
+//! resolve at a fixed intervals in future
+
+use std::io;
+use std::time::{Duration, Instant};
+
+use futures::{Poll, Async};
+use futures::stream::{Stream};
+
+use reactor::{Remote, Handle};
+use reactor::timeout_token::TimeoutToken;
+
+/// A stream representing notifications at fixed interval
+///
+/// Intervals are created through the `Interval::new` or
+/// `Interval::new_at` methods indicating when a first notification
+/// should be triggered and when it will be repeated.
+///
+/// Note that timeouts are not intended for high resolution timers, but rather
+/// they will likely fire some granularity after the exact instant that they're
+/// otherwise indicated to fire at.
+pub struct Interval {
+    token: TimeoutToken,
+    next: Instant,
+    interval: Duration,
+    handle: Remote,
+}
+
+impl Interval {
+    /// Creates a new interval which will fire at `dur` time into the future,
+    /// and will repeat every `dur` interval after
+    ///
+    /// This function will return a future that will resolve to the actual
+    /// interval object. The interval object itself is then a stream which will
+    /// be set to fire at the specified intervals
+    pub fn new(dur: Duration, handle: &Handle) -> io::Result<Interval> {
+        Interval::new_at(Instant::now() + dur, dur, handle)
+    }
+
+    /// Creates a new interval which will fire at the time specified by `at`,
+    /// and then will repeat every `dur` interval after
+    ///
+    /// This function will return a future that will resolve to the actual
+    /// timeout object. The timeout object itself is then a future which will be
+    /// set to fire at the specified point in the future.
+    pub fn new_at(at: Instant, dur: Duration, handle: &Handle)
+        -> io::Result<Interval>
+    {
+        Ok(Interval {
+            token: try!(TimeoutToken::new(at, &handle)),
+            next: at,
+            interval: dur,
+            handle: handle.remote().clone(),
+        })
+    }
+}
+
+impl Stream for Interval {
+    type Item = ();
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<Option<()>, io::Error> {
+        // TODO: is this fast enough?
+        let now = Instant::now();
+        if self.next <= now {
+            self.next = next_interval(self.next, now, self.interval);
+            self.token.reset_timeout(self.next, &self.handle);
+            Ok(Async::Ready(Some(())))
+        } else {
+            self.token.update_timeout(&self.handle);
+            Ok(Async::NotReady)
+        }
+    }
+}
+
+impl Drop for Interval {
+    fn drop(&mut self) {
+        self.token.cancel_timeout(&self.handle);
+    }
+}
+
+/// Converts Duration object to raw nanoseconds if possible
+///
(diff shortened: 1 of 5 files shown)


alexcrichton (commented):
Thanks for the PR! This looks pretty good to me, but can it also remove `TimeoutToken::when` while you're at it? Historically it was added as the precise time a timeout fired could change due to rounding, but now we don't do that any more.

tailhook:
Moved `when` to the `Timeout` rather than `TimeoutToken`. Is this what you meant? (I could remove time check at all but I don't think it's a good idea)

I've also made `reactor::Inner::add_timeout` return just `usize` as it never fails any more.

carllerche:
My main question is, what is the best behavior for scheduling the next timeout and what happens if the consumer falls behind?

Should the next timeout be scheduled when the current one has fired. If a consumer falls behind, should the interval stream be ready immediately for each missed interval? Should it only fire once for missed intervals? Should it provide the number of missed intervals?

I also think that we should figure out how to unify with `tokio-timer` and try to avoid deviating too much.

The related tokio-timer issue: tokio-rs/tokio-timer#6

tailhook:
> My main question is, what is the best behavior for scheduling the next timeout and what happens if the consumer falls behind?

Yes, this is a good question. In my opinion, there are two use cases:
1. When  you don't care. I.e. you send a notification to a monitoring system, of a ping to other peer. If consumer falls behind you don't want ten pings to be sent at the same time
2. If you really care, say your implementation renders an animation you have time bookeeping in your code anyway. 

Also, I have no code to prove, but I believe that making bookeeping of missing timeouts is easier than other way around (i.e. calculate whether to skip this event on the stream) because you don't know whether another timeout will happen shortly.

So current implementation tries to avoid constant drift of the interval unless consumer falls behind (if I understand streams correctly). In that case, it schedules next timeout "starting from now". 

> Should it provide the number of missed intervals?

Well, for those cases where it matters, say you're implementing an animation, you shouldn't rely on number of missed intervals provided by interval stream. You can always calculate it by `(new_time - old_type)/interval`. And that would be more precise for the cases where you have some other things between stream firing a timeout and the code accounting animation.

src/reactor/mod.rs
```diff
@@ -460,14 +466,24 @@ impl Inner {
...
 
     fn update_timeout(&mut self, token: usize, handle: Task) -> Option<Task> {
         debug!("updating a timeout: {}", token);
         self.timeouts[token].1.block(handle)
     }
 
+    fn reset_timeout(&mut self, token: usize, at: Instant) {
+        let pair = &mut self.timeouts[token];
+        if let Some(slot) = pair.0.take() {
+            self.timer_heap.remove(slot);
```
alexcrichton, line 483: Could you add a TODO here to avoid remove + push and instead just do one sift of the heap? In theory we could update it in place and then do the percolation as necessary

src/reactor/interval.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    fn poll(&mut self) -> Poll<Option<()>, io::Error> {
+        // TODO: is this fast enough?
+        let now = Instant::now();
+        if self.next <= now {
+            // Try to avoid drift if possible
+            self.next += self.interval;
+            // But if we are too far in future just start interval from now
+            if self.next < now {
+                self.next = now + self.interval;
```
alexcrichton: I'd personally expect that all timeouts would be an integral number of units of the interval from the start point, but it looks like this is basically resetting the start point to the current point in time? That seems like you could have an extra long timer instead of a regular period?

Perhaps this should be the moral equivalent of:

``` rust
while self.next <= now {
    self.next += self.interval;
}
```

alexcrichton:
@carllerche good points! I'm fine with the "fast forward" approach taken here and it seems like we can shake  out the specifics in tokio-timer and then inherit the semantics when that's used to build the support here.

tailhook:
> I'd personally expect that all timeouts would be an integral number of units of the interval from the start point, but it looks like this is basically resetting the start point to the current point in time?

Yes, that was made to have a simpler implementation. Take a look at current one. It solved the problem at the expense of more calculations (they are weird, but I've tried to document them as much as possible)

> That seems like you could have an extra long timer instead of a regular period?

No. The largest interval it could get is the specified one, sans the inaccuracy of event dispatching. The only deficiency that implementation had is in the case you're trying to align your events on a second boundary (i.e. connected to the real-time), or it could change a difference in time when there are two intervals with the same interval value but different `at` (start time).  Note that the first use case (aligning events on a second boundary) doesn't work too well too, because `Instant` may drift arbitrarily comparing to the wall clock time anyway.

alexcrichton:
@tailhook was perhaps a push forgotten? Or is what's here the most recent version?

tailhook:
> @tailhook was perhaps a push forgotten? Or is what's here the most recent version?

Yes. Now it's up to date.

src/reactor/interval.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/// Multiply duration by u64
+///
+/// Rust only supports multiplying Duration by u32, so we have to
+/// multiply it manually.
+///
+/// Note: this implementatin crashes on multiplier bigger than (1 << 63) - 1,
+/// but we can't get a number this large in the next_interval function.
+fn duration_multiply(dur: Duration, mult: u64) -> Duration {
```
alexcrichton: Perhaps we could conservatively assume for now that the `mult` here always fits in a `u32`? I'd have to think pretty hard about the correctness here...

tailhook:
> Perhaps we could conservatively assume for now that the mult here always fits in a u32? I'd have to think pretty hard about the correctness here...

Done that. But for the reference:
1. Code was not that complex, just split u64 into high bits and low bits doing multiply in turn
2. Technically it's possible to trigger the assertion (> u32) having duration of 1 ns and being slow for 4 seconds
3. While interval of 1ns is unrealistic, with 1ms and 100ns interval you could trigger a but by freezing application for ~42 days and ~4 days respectively

I.e. overflow is very unlikely to happen often, but calculations certainly show that this is not the totally unrealistic situation which could be triggered by the combination of poorly chosen intervals and STOP/CONT signals, CRIU, or slow file system.

On the other hand, it's probably a good idea to implement u64 multiplication for a Duration itself.

alexcrichton:
Looks great to me, thanks!

PR #58, opened 2016-10-14
fix(examples): Fix typos in chat example

oberien:
Rephrase a sentence to not contain the word `join` when we are actually using
`select` to prevent confusion.


alexcrichton:
Thanks!

PR #60, opened 2016-10-17
[WIP] spawn_many function

tailhook:
This work in progress pull request implements `handle.spawn_many` function, which is very close to what `Stream::buffered()` does except it spawns all the futures as separate tasks.

In this implementation, it works but does not limit the number of tasks spawned simultaneously. This is just to show you the usage.

It simplifies example:

```
    let done = socket.incoming().for_each(move |(socket, addr)| {
        let pair = futures::lazy(|| futures::finished(socket.split()));
        let amt = pair.and_then(|(reader, writer)| copy(reader, writer));

        // Once all that is done we print out how much we wrote, and then
        // critically we *spawn* this future which allows it to run
        // concurrently with other connections.
        let msg = amt.map(move |amt| {
            println!("wrote {} bytes to {}", amt, addr)
        }).map_err(|e| {
            panic!("error: {}", e);
        });
        handle.spawn(msg);

        Ok(())
    });
```

Into:

```
    handle.spawn_many(100, socket.incoming().map(move |(socket, addr)| {
        futures::lazy(|| futures::finished(socket.split()))
        .and_then(|(reader, writer)| copy(reader, writer))
        .map(move |amt| {
            println!("wrote {} bytes to {}", amt, addr)
            })
        .map_err(|e| {
            panic!("error: {}", e);
            })
    }).map_err(|_| ()));
```

And also limits number of simultenous connections (eventually, when fully implemented).

I think in proposed form it looks quite cool:
1. You don't have to care about the result of `for_each`, as the stream is infinite anyway
2. The number of tasks is limited, which should be implemented in any decent server
3. It works for any stream
4. It allows nice error reporting (every error is required to converge to `()`, so you ought to log and drop it)
5. It doesn't require cloning a handle from user point of view which is quite common point of problems for newcomers

Is this going to be accepted? Do you like this interface or should it be changed? Thoughts?

### DIFF
--- a/examples/spawn_many.rs
+++ b/examples/spawn_many.rs
@@ -1,0 +1,47 @@
+//! An echo server that just writes back everything that's written to it.
+
+extern crate env_logger;
+extern crate futures;
+extern crate tokio_core;
+
+use std::env;
+use std::net::SocketAddr;
+
+use futures::{Future, empty};
+use futures::stream::Stream;
+use tokio_core::io::{copy, Io};
+use tokio_core::net::TcpListener;
+use tokio_core::reactor::Core;
+
+fn main() {
+    env_logger::init().unwrap();
+    let addr = env::args().nth(1).unwrap_or("127.0.0.1:8080".to_string());
+    let addr = addr.parse::<SocketAddr>().unwrap();
+
+    // Create the event loop that will drive this server
+    let mut l = Core::new().unwrap();
+    let handle = l.handle();
+
+    // Create a TCP listener which will listen for incoming connections
+    let socket = TcpListener::bind(&addr, &handle).unwrap();
+
+    // Once we've got the TCP listener, inform that we have it
+    println!("Listening on: {}", addr);
+
+    // Pull out the stream of incoming connections and then for each new
+    // one spin up a new task copying data.
+    //
+    // We use the `io::copy` future to copy all data from the
+    // reading half onto the writing half.
+    handle.spawn_many(1, socket.incoming().map(move |(socket, addr)| {
+        futures::lazy(|| futures::finished(socket.split()))
+        .and_then(|(reader, writer)| copy(reader, writer))
+        .map(move |amt| {
+            println!("wrote {} bytes to {}", amt, addr)
+            })
+        .map_err(|e| {
+            panic!("error: {}", e);
+            })
+    }).map_err(|_| ()));
+    l.run(empty::<(), ()>()).unwrap();
+}
--- a/src/reactor/mod.rs
+++ b/src/reactor/mod.rs
@@ -14,6 +14,7 @@ use std::time::{Instant, Duration};
 
 use futures::{self, Future, IntoFuture, Async};
 use futures::task::{self, Unpark, Task, Spawn};
+use futures::stream::Stream;
 use mio;
 use slab::Slab;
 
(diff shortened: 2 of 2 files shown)


alexcrichton:
Ideally I'd love to see this implemented as a combinator as I think that'd also enable this sort of power to be built externally. In theory this should be possible with the [`buffered_unordered`](http://alexcrichton.com/futures-rs/futures/stream/trait.Stream.html#method.buffer_unordered) method on streams, right?

tailhook:
> Ideally I'd love to see this implemented as a combinator as I think that'd also enable this sort of power to be built externally. In theory this should be possible with the buffered_unordered method on streams, right?

I don't see how you expect this to work. `buffered_unordered` does just the same, but it doesn't scale. You can't make `stream.buffered_unordered(1_000_000)`. But you can do that even with this example implementation of `spawn_many`.

I can imagine something like `stream.buffered_unordered_spawn(1_000_000, |f| handle.spawn(f))`, but I'm not sure it's possible to implement in such a generic way (or do you have any ideas?) I think the exact backpressure mechanism is specific to tokio_core.

It _can_ be built externally as a function `spawn_many(1_000_000, handle, stream)` but I feel it's very basic building block that should be out of the box.

I mean showing  `spawn` with unlimited number of tasks in examples and providing no upgrade path leads to the fact that [everyone](https://github.com/tokio-rs/tokio-proto/blob/0bfe489f2bd97307024df8f0c7a7ed112f8253e5/src/server.rs#L73) [uses](https://github.com/icorderi/tokio-netstring/blob/master/src/service.rs#L43) [exactly same](https://github.com/popravich/minihttp/blob/master/src/lib.rs#L117) bad code. While it can be fixed directly in `service::listen`, I think more basic solution which works on arbitrary streams is better.

carllerche:
IMO if this is a combinator targeted specifically for accepting sockets, it should live as a helper vs. including it on handle. At the very least it can start as a helper and get promoted as it evolves.

carllerche:
That being said, having a way to cap incoming connections seems good, but in practice it probably ideally wouldn't be based on a fixed number and instead somehow fall out of a back pressure system.

ghost:
This is really cool combinator!

Current implementation of `buffer_unordered` indeed does not scale well with
the number of concurrent futures, as it may have to poll every single
underlying future to find which one is ready.

Though, the `task::with_unpark_event` could be used to quickly determine which
future should be polled from. Effectively providing a very similar functionality.
For one place where `with_unpark_event` is already used see unix version of
tokio-curl: https://github.com/tokio-rs/tokio-curl/blob/master/src/unix.rs

tailhook:
>  IMO if this is a combinator targeted specifically for accepting sockets, it should live as a helper vs. including it on handle. At the very least it can start as a helper and get promoted as it evolves.

A function `tokio_core::reactor::spawn_many` would be fine?

> That being said, having a way to cap incoming connections seems good, but in practice it probably ideally wouldn't be based on a fixed number and instead somehow fall out of a back pressure system.

Well, I think even if there is a flexible backpressure, a hard limit is good too. While actual hard limit is `EMFILE` usually we want to leave some file descriptors for processing requests otherwise we could have a deadlock (or a livelock).

But maybe you can elaborate on this topic a litle bit.

> Though, the task::with_unpark_event  lcould be used to quickly determine which future should be polled from. 

Well, I'm not sure, is it really as scalable as `spawn()` ? Also, I'm still not sure how it works... At a glance, it requires cooperation with the implementation of every Future to work efficienly.

alexcrichton:
It sounds like whether this is a combinator or not boils down to whether `buffered` or `buffered_unordered` can be used. As @tmiasko pointed out `with_unpark_event` is not currently used but could be used to efficiently implement this with `buffered_unordered`.

In that sense I don't see a reason why this _can't_ be just another combinator? We could perhaps provide it by default in tokio-core to start, but I could imagine that it'd want to migrate down to the futures layer eventually as well.

alexcrichton:
Ok with https://github.com/alexcrichton/futures-rs/pull/216 landed and the numbers you posted there I'm going to close this, but we can continue to explore our options here if that's not sufficient!

PR #64, opened 2016-10-20
Move Framed from tokio-proto to core

alexcrichton:
This commit extracts the concrete implementation of `FrameIo` in tokio-proto to
tokio-core under the name `EasyFramed`. This extraction is accompanied with a
new `EasyBuf` buffer type to work with when parsing types.

The purpose of this movement is to provide a clear and easy entry point at the
`FramedIo` layer for those who need it. Eventually these buffer types will get
replaced or moved to the `bytes` crate, but in the interest of an 0.1 release
and remaining backwards compatible with the tokio-core 0.1 release this is
adding a separate module.

### DIFF
--- a/src/easy.rs
+++ b/src/easy.rs
@@ -1,0 +1,438 @@
+//! A module for working with "easy" types to interact with other parts of
+//! tokio-core.
+//!
+//! This module contains a number of concrete implementations of various
+//! abstractions in tokio-core. The contents of this module are not necessarily
+//! production ready but are intended to allow projects to get off the ground
+//! quickly while also showing off sample implementations of these traits.
+//!
+//! Currently this module primarily contains `EasyFramed`, a struct which
+//! implements the `FramedIo` trait in `tokio_core::io`. This structure allows
+//! simply defining a parser (via the `Parse` trait) and a serializer (via the
+//! `Serialize` trait) and transforming a stream of bytes into a stream of
+//! frames. Additionally the `Parse` trait passes an `EasyBuf`, another type
+//! here, which primarily supports `drain_to`, to extract bytes without copying
+//! them.
+//!
+//! For more information see the `EasyFramed` and `EasyBuf` types.
+
+use std::io;
+use std::ops::{Deref, DerefMut};
+use std::sync::Arc;
+
+use futures::{Async, Poll};
+
+use io::{Io, FramedIo};
+
+/// A reference counted buffer of bytes.
+///
+/// An `EasyBuf` is a representation of a byte buffer where sub-slices of it can
+/// be handed out efficiently, each with a `'static` lifetime which keeps the
+/// data alive. The buffer also supports mutation but may require bytes to be
+/// copied to complete the operation.
+pub struct EasyBuf {
+    buf: Arc<Vec<u8>>,
+    start: usize,
+    end: usize,
+}
+
+/// An RAII object returned from `get_mut` which provides mutable access to the
+/// underlying `Vec<u8>`.
+pub struct EasyBufMut<'a> {
+    buf: &'a mut Vec<u8>,
+    end: &'a mut usize,
+}
+
+impl EasyBuf {
+    /// Creates a new EasyBuf with no data and the default capacity.
+    pub fn new() -> EasyBuf {
+        EasyBuf::with_capacity(8 * 1024)
+    }
+
+    /// Creates a new EasyBuf with `cap` capacity.
+    pub fn with_capacity(cap: usize) -> EasyBuf {
+        EasyBuf {
+            buf: Arc::new(Vec::with_capacity(cap)),
+            start: 0,
+            end: 0,
+        }
+    }
+
+    /// Changes the starting index of this window to the index specified.
+    ///
+    /// Returns the windows back to chain multiple calls to this method.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if `start` is out of bounds for the underlying
+    /// slice or if it comes after the `end` configured in this window.
+    fn set_start(&mut self, start: usize) -> &mut EasyBuf {
+        assert!(start <= self.buf.as_ref().len());
+        assert!(start <= self.end);
+        self.start = start;
+        self
+    }
+
+    /// Changes the end index of this window to the index specified.
+    ///
+    /// Returns the windows back to chain multiple calls to this method.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if `end` is out of bounds for the underlying
+    /// slice or if it comes after the `end` configured in this window.
+    fn set_end(&mut self, end: usize) -> &mut EasyBuf {
+        assert!(end <= self.buf.len());
+        assert!(self.start <= end);
+        self.end = end;
+        self
+    }
+
+    /// Returns the number of bytes contained in this `EasyBuf`.
+    pub fn len(&self) -> usize {
+        self.end - self.start
+    }
+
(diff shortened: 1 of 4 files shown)


alexcrichton:
cc @rrichardson and https://github.com/tokio-rs/tokio-core/pull/35

This is somewhat similar but I still think orthogonal. We're currently thinking of delaying `bytes` integration until 0.2, so we may hold off on #35 for now.

tailhook:
While this is definitely much better than `tokio-proto::Framed` for tutorials, I'm not sure what's the point in adding it to a tokio-core.

I ever wondered why `FramedIo` is in tokio-core too. It isn't used in this crate. And is only useful with transport implementations in tokio-proto.

While a first it might seem to be a good idea to implement your own "transports" based on FramedIo, in my practice it ends up either reimplementation of tokio-proto for no good reason (performance? backpressure?) or has some peculiarities like in [our http implementation](https://github.com/popravich/minihttp) which requires handing off a socket to disk thread and returning back (it has more differences, this one is just easier to explain), so FrameIo doesn't work here.

Do you know any projects that use FramedIo but are not using tokio-proto? What are the use cases?

src/easy.rs
```diff
@@ -0,0 +1,439 @@
...
+    fn read(&mut self) -> Poll<Self::Out, io::Error> {
+        loop {
+            // If the read buffer has any pending data, then it could be
+            // possible that `parse` will return a new frame. We leave it to
+            // the parser to optimize detecting that more data is required.
+            if self.is_readable {
+                if self.eof {
+                    if self.rd.len() == 0 {
+                        // TODO: seems like this should return `None`?
+                        return Ok(Async::NotReady)
```
ghost: I kinda like how this detects that there are no bytes outside the parser and
allows `Parser::done` to return plain result. 

Though, it has the downside that it does not allow parser to manufacture ending
frame out of thin air, because there is no guarantee that `Parser::done` is
called even once after the end of stream. 

Consider situation where all bytes are already consumed and some time later
peer closes connection for writing. The parser will not be called because there
is no new data to consume.

src/io/mod.rs
```diff
@@ -131,6 +131,9 @@ pub trait Io: io::Read + io::Write {
 /// value that indicates "would block" the current future's task is arranged to
 /// receive a notification when the method would otherwise not indicate that it
 /// would block.
+///
+/// For a sample implementation of `FramedIo` you can take a look at the
+/// `EasyFramed` type in the `easy` module of htis crate.
```
ghost, line 136: s/htis/this

tailhook:
Additional thought: keeping it in tokio-proto, or any other crate besides `tokio-core` allows different protocols use different versions of Framed/FramedIo, whatever. While `tokio-core` version must be consistent for all the libraries used.

alexcrichton:
@tailhook 

In some sense tokio-core basically doesn't use any of its own abstractions (e.g. `Io`, `split`, `TcpStream`, etc), so I don't think that's a great way to measure whether it's useful here or not. Instead tokio-core is laying the foundation for the rest of the tokio ecosystem. For example this is basically a case where it's saying if you have a framed protocol then you should probably implement this trait, and if you want to work with framed protocols you should work with this trait.

> While a first it might seem to be a good idea to implement your own "transports" based on FramedIo, in my practice it ends up either reimplementation of tokio-proto for no good reason

This seems bad! If these concerns are cropping up at lower layers then the constraints definitely need to get pushed down to help design the lower layers as well. A constraint from one scenario is likely a constraint that others are worried about as well!

Do you have some concrete examples of where `FramedIo` as a trait fell down?

> Do you know any projects that use FramedIo but are not using tokio-proto? What are the use cases?

Not currently, but in terms of entry points `FramedIo` is a great way to enter into the ecosystem. We could also add a dead-simple future which takes a `FramedIo`, a service of what to do with a frame, and then runs that to completion. E.g. that's just a server that processes at most one request at a time but can process many per connection. This ability to jump from a `TcpStream` to a `Service` with only defining a parser can be quite handy for getting up and running.

---

Also note that this is just an implementation (e.g. explicitly labeled **Easy** to not indicate it's useful for everything) in tokio-core for ease. This is not intended to be the end-all-be-all of implementations, and that's why it's primarily the `FramedIo` trait and not the `EasyFramedIo` struct.

src/easy.rs
alexcrichton: Interesting! This made me want to change the trait to return `Poll<Option<Frame>, ...>` so signal EOF. What protocols though have a frame at the end of a stream when there are no bytes available?

carllerche:
The primary reason to provide the `FramedIo` trait in core is to establish a common trait across the ecosystem that can be used to implement "framing". Standardizing "framing" enables splitting up the various steps into individual, reusable components. This is an idea that has a lot of prior art and has been proven successful (netty, wangle, ....)

For example, a protobuf `FramedIo` implementation can limit itself to only doing the protobuf serialization / deserialization step. It can be initialized w/ `T: FramedIo<In = impl AsRef<[u8]>, Out = impl AsRef<[u8]>>` which handles the details of delineating the serialized messages.

So, you could then do:

``` rust
let len_delimited = LengthDelimited::new(io);
let protobuf = ProtobufFramedIo::new(len_delimted);
```

In short, `FramedIo` provides the basis on which such an ecosystem of components can be built, which is why it lives in `core`

The **only** thing that `tokio-proto` provides is the glue layer between `FrameIo` and `Service`. It's only supposed to be an implementation detail that client / server implementors can choose to use and should stay entirely encapsulated in Tokio based libraries.

```
```

rrichardson:
:+1:

src/easy.rs
ghost: The idea of making this more stream-like with EOF expressed at type level is
interesting. Currently tokio-proto uses `Frame::Done` to indicate that
(`Frame::Done` doesn't carry any payload so it is likely to be constructed
without consuming any input bytes, just from the fact that input stream has
ended), but you need to understand specific frame type to know whether it is
EOF or not, otherwise all frames are completely opaque.

Though, if we were to use `Poll<Option<Frame>, ...>`, what `Async::NotReady`
would mean (**edit**: when returned from `done`)? Should you continue polling
from it, or not? (could a parser have a future inside that has to complete
before another frame is returned?)

alexcrichton: Oh note that `done` returns a `Result` where `parse` returns `Poll`. That is, at `done`, you can never be `NotReady`.

alexcrichton:
@carllerche I'm curious what you think about [this comment](https://github.com/tokio-rs/tokio-core/pull/64/files#diff-9840ff82017e5c1ce3b8ca1398e902a1R363), specifically eventually changing the signature of `Parse::parse` to:

``` rust
fn read(&mut self) -> Poll<Option<Self::Out>, io::Error>;
```

Right now if we hit EOF and the internal buffer has 0 bytes the implementation here just returns `NotReady`, but it technically needs to propagate upwards "the stream is done"

carllerche:
Yeah, seems good to me. The argument for makes sense.

SirVer:
> Do you know any projects that use FramedIo but are not using tokio-proto?

Just for posterity: I have trouble using tokio-proto with my project - since I have trouble using tokio-service. 

My protocol is chunked in frames. Each frame begins with 2 bytes of payload length and the payload. This could be wonderfully expressed through `tokio_proto`, if it were not for the dependency on the `tokio_service` crate.

My server and client hold a long connection and occasionally one of the two will send a package and might or might not want a reply back. Especially, there is no request -> response mechanism in play that the Service trait abstracts over. Or maybe I am doing it wrong.

Issue #66, opened 2016-10-20
Consider splitting FramedIo -> FramedRead + FramedWrite


carllerche:
... or maybe `ReadFrame` + `WriteFrame`

relates to #62

cc @Ralith

aturon:
As part of this, reconsider the use-case for `poll_read`. Ideally, the split here is something like a `Future` plus a `Sink` (if/when the latter becomes part of the futures crate).

aturon:
I don't think this is relevant any longer.

carllerche:
To clarify, this basically happened by using `Sink + Stream`

PR #82, opened 2016-10-25
Polish tokio-core in prep for overall 0.1 release

aturon:
This commit makes a few tweaks to the new `easy` module:
- ~~Rename `Parse` to `Deserialize`, for symmetry with the `Serialize`
  trait.~~
- Rename `Parse` to `Decode` and `Serialize` to `Encode`.
- Don't use `Poll` for the `deserialize` method; we prefer to reserve
  that type for actual aync events, and in particular for a `NotReady`
  result to imply that some task scheduling has taken place. Instead,
  use an internal `Option`.

cc @carllerche @alexcrichton

### DIFF
--- a/src/easy.rs
+++ b/src/easy.rs
@@ -8,11 +8,11 @@
 //!
 //! Currently this module primarily contains `EasyFramed`, a struct which
 //! implements the `FramedIo` trait in `tokio_core::io`. This structure allows
-//! simply defining a parser (via the `Parse` trait) and a serializer (via the
-//! `Serialize` trait) and transforming a stream of bytes into a stream of
-//! frames. Additionally the `Parse` trait passes an `EasyBuf`, another type
-//! here, which primarily supports `drain_to`, to extract bytes without copying
-//! them.
+//! simply defining a decoder (via the `Decode` trait) and a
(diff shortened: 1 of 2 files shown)


tailhook:
Looks good to me.

And if we are simplifying this thing, are there any good aliases for `In` and `Out` ? I mean it's often messed up because `In` looks like something that is read from the socket, but it's the opposite. Maybe `Source`, `Target` are good alternatives?

alexcrichton:
👍

carllerche:
Since this touches naming, do we want to consider using `Encode` / `Decode` instead? It is shorter and this is the terminology I've seen in the past.

alexcrichton:
Hm yeah if I had to pick two names I like Encode/Decode, nice, short, accepted, etc.

aturon:
@tailhook I do think we want to reconsider those names, but I'm trying to avoid any breakage against what's been published in 0.1 so far. The `FramedIo` trait is likely to be [re-thought entirely](https://github.com/tokio-rs/tokio-core/issues/66) in 0.2.

@carllerche @alexcrichton Yes, `encode`/`decode` is even better; I'll change it.

PR #88, opened 2016-11-02
Add UDP echo server example.

dpc:


dpc:
I am just toying, and I thought it might be useful. Please let me know in case this is wrong, or not the best approach or something.

examples/echo-udp.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl Future for Server {
+    type Item = ();
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<(), io::Error> {
+        loop {
+            if let Some((size, peer)) = self.to_send.take() {
+                match self.socket.send_to(&self.buf[..size], &peer) {
+                    Err(e) => {
```
alexcrichton, line 39: I'm not super familiar with UDP, but presumably this can return "would block", right? In that case we should handle that specifically, right, and return `NotReady`?

alexcrichton:
Nice!

Out of curiosity, is there a good way to interact with this on the command line? I typically use `nc` or `telnet` for TCP but haven't toyed much with UDP in the past...

dpc:
```
echo -n "hello" | nc -4u  localhost 9999
```

examples/echo-udp.rs
dpc: Right. Will fix.

dpc: Done.

alexcrichton: Hm was a push forgotten? (looks the same online at least)

alexcrichton: Oh! nevermind

PR #90, opened 2016-11-05
WIP: Add some benchmarks.

dpc:
To be moved into separate repo.

### DIFF
--- a/benches/latency.rs
+++ b/benches/latency.rs
@@ -18,17 +18,19 @@ use test::Bencher;
 use std::thread;
 use std::time::Duration;
 
+use futures::stream::Stream;
+
 /// UDP echo server
-struct Server {
+struct EchoServer {
     socket : UdpSocket,
     buf : Vec<u8>,
     to_send : Option<(usize, SocketAddr)>,
     stop : Arc<AtomicBool>,
 }
 
-impl Server {
+impl EchoServer {
     fn new(s : UdpSocket, stop : Arc<AtomicBool>) -> Self {
-        Server {
+        EchoServer {
             socket: s,
             to_send: None,
             buf: vec![0u8; 1600],
@@ -37,7 +39,7 @@ impl Server {
     }
 }
 
-impl Future for Server {
+impl Future for EchoServer {
     type Item = ();
     type Error = io::Error;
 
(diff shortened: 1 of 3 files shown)


dpc:
Unfortunately, UDP has too mach variance to be really useful. I guess I'll try with `channel` instead.

```
test udp_echo_latency ... bench: 19,681 ns/iter (+/- 12,347)
```

alexcrichton:
Thanks! I wonder if perhaps the mio benchmarks should go into mio itself?

dpc:
I have fixed the tests and added a better one: channel, which should not be affected by networking layer, firewall overhead etc. Current test code is not prettiest, but it seems to get the job done. Please let me know if you see any fundamental problems.

```
test channel_latency  ... bench:       6,049 ns/iter (+/- 532)
test udp_echo_latency ... bench:       8,182 ns/iter (+/- 381)
```

Now that I can actually measure the response time, I can look into what influences it, and maybe optimize.

benches/latency.rs
```diff
@@ -0,0 +1,218 @@
...
+        let socket = tokio_core::net::UdpSocket::bind(&server_addr, &handle).unwrap();
+
+        let server = EchoServer::new(socket, stop);
+
+        l.run(server).unwrap();
+    });
+
+    // TODO: More reliable way to bind server socket and start server
+    // first
+    thread::sleep(Duration::from_millis(100));
```
alexcrichton: Could a channel be used instead of this to send a message as soon as the socket is bound?

benches/latency.rs
```diff
@@ -0,0 +1,218 @@
...
+        ChanServer {
+            rx: rx,
+            tx: tx,
+            buf: None,
+            stop: stop,
+        }
+    }
+}
+
+impl Future for ChanServer {
```
alexcrichton, line 81: This may be replace-able with the newly-minted `Sink::send_all` function in the futures crate

benches/latency.rs
```diff
@@ -0,0 +1,218 @@
...
+        client.send_to(&buf, &server_addr).unwrap();
+        let _ = client.recv_from(&mut buf).unwrap();
+    }
+
+    b.iter(|| {
+        client.send_to(&buf, &server_addr).unwrap();
+        let _ = client.recv_from(&mut buf).unwrap();
+    });
+
+    // Stop the server; TODO: Use better method
```
alexcrichton: Could this use a oneshot, and the server is waiting on that?

benches/latency.rs
```diff
@@ -0,0 +1,218 @@
...
+                Ok(futures::Async::Ready(Some(t))) => { self.buf = Some(t) },
+                Ok(futures::Async::NotReady) => return Ok(futures::Async::NotReady),
+                Err(e) => return try_nb!(Err(e)),
+            }
+        }
+    }
+}
+#[bench]
+fn udp_echo_latency(b: &mut Bencher) {
+    let server_addr= "127.0.0.1:7398".to_string();
```
alexcrichton: Instead of using hardcoded ports, could this bind to port 0 to allow it to use any free port on the system?

benches/latency.rs
```diff
@@ -0,0 +1,218 @@
...
+    // TODO: More reliable way to bind server socket and start server
+    // first
+    thread::sleep(Duration::from_millis(100));
+
+    let mut rx_iter = out_rx.wait();
+
+    // warmup phase; for some reason initial couple of
+    // rounds is much slower
+    //
+    // TODO: Describe the exact reasons; caching? branch predictor? lazy closures?
```
alexcrichton, line 179: Was this required even for channels? I could kinda believe it for UDP but having that much variance for channels seems odd.

benches/latency.rs
```diff
@@ -0,0 +1,218 @@
...
+    type Item = ();
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<(), io::Error> {
+        loop {
+            if self.stop.load(Ordering::SeqCst) {
+                return Ok(futures::Async::Ready(()))
+            }
+
+            if let Some((size, peer)) = self.to_send.take() {
```
alexcrichton: Presumably this could use `.as_ref()`, and then `try_nb!` could be used below to elide the whole match statement?

benches/latency.rs
dpc: Yes. I've actually added it to make sure server had time to start up and can respond immediately. I'll double check and investigate more.

benches/latency.rs
dpc: Definitely. Will do.

benches/latency.rs
dpc: Yes. Though I'll have to send back the actual channel to the client thread, so I started with hardcoded version.

alexcrichton: Sure but a message needs to be sent anyway as soon as the socket is bound, so this'd just be precisely that message

dpc:
Thanks for feedback. I'll work on it.

dpc:
I think I addressed everything, except `Sink::send_all` as it doesn't seem to be released yet (or did I miss it?).

The "warmup" investigation pointed to the edge of 3-5 warmup runs being necessary for `channel` and usually just `1` for UDP. But unfortunately I still don't understand why.

alexcrichton:
Thanks!

PR #91, opened 2016-11-05
Simply echo example: split() can now happen on the main task.

dwrensha:
https://github.com/tokio-rs/tokio-core/commit/614887b8c15943b4afae494af83a815058f128a3 means that this will no longer panic.


alexcrichton:
yay!

PR #93, opened 2016-11-08
Depend directly on git version of futures-rs

danburkert:
Cargo replace directives are not meant for libraries, since they are not
transitively inherited by the application.


alexcrichton:
Yeah I'm not really sure of the best way to deal with these problems, but I guess this is as good as any in the interim!

PR #95, opened 2016-11-08
Refactor framing to use Streams and Sinks

aturon:
- Gets rid of `easy` module, instead providing framing support directly
  in the `io` module.

- ~In particular, adds framing adapters directly to the `Io` trait. These
  adapters give you Stream/Sink objects, either as separate objects or
  in unified ("duplex") form.~ Now just a single framing adapter, which gives you a combined object you can split.

- Deprecates the `FramedIo` trait; that's now just Stream + Sink.

- Updates the line framing test to use the stream/sink combinators.

### DIFF
--- a/src/io/frame.rs
+++ b/src/io/frame.rs
@@ -17,12 +17,14 @@
 //! For more information see the `EasyFramed` and `EasyBuf` types.
 
 use std::io;
+use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 
-use futures::{Async, Poll};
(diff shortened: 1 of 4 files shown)


carllerche:
IMO having the entry point be `framed(io)` (+ an alternate `framed_duplex`) vs. `framed().split()` makes the entire system more complex as it implicitly introduces the concept of "splitting".

One will end up w/ two separate handles and if having a joint type is required, then learning about "splitting" will have to happen in a round about way vs. having the "fused" type an discovering the `split` fn.

Examples can all include `split` as well, which makes the pattern obvious.

aturon:
On reflection, I think that @carllerche is right, and I've updated the PR accordingly.

PR #97, opened 2016-11-09
Correct client invocation example

k4rtik:


alexcrichton:
Thanks! If you run `nc` though doesn't it allow you to keep typing messages?

k4rtik:
@alexcrichton You are right, I just needed to add `-4` to get it to work. Thanks.

alexcrichton:
Thanks!

PR #98, opened 2016-11-10
export framed function from frame module

cmsd2:
The frame module is private so the framed constructor function isn't visible.
This makes it public by explicitly exporting it from the io module.


aturon:
Thanks for the PR, but this is already available through the `Io` trait, using the `framed` method.

aturon:
(See https://github.com/cmsd2/tokio-core/blob/445a9443ea323299515d0f97862ccaeb7d5623a9/src/io/mod.rs#L132)

PR #100, opened 2016-11-12
Proxy example and changes motivated by it: take 2

dwrensha:
This in an iteration on https://github.com/tokio-rs/tokio-core/pull/28.

Adds `ReadHalf::with()` and `WriteHalf::with()` methods to allow `shutdown()` to be called.


alexcrichton:
Nice!

I'm 👍 on all these changes, we'll just have to wait for 0.2 unfortunately due to "fixing" the `copy` combinator. Thanks @dwrensha!

alexcrichton:
Ok I've added this example in https://github.com/tokio-rs/tokio-core/commit/a8e09c5608381907e8597609387cfff83901467d and otherwise the move to `tokio-io` has obsoleted the need for a number of these changes in this repository so I'm going to close this. Let me know though if anything looks awry!

PR #102, opened 2016-11-14
Reintroduce "decoder" and "encoder" for Decode and Encode, merging them into Codec

aturon:
A previous commit refactored such that `Encode` and `Decode` are
implemented directly on the types being encoded or decoded. This was
thought to be less expressive but more convenient than having a separate
notion of a (stateful) encoder or decoder.

However, there are certain situations where the approach is just too
limiting: you're required to implemented `Decode` and `Encode` for types
you don't "own" and can't newtype.

This commit moves back to a setup where `Self` represents the
encoder/decoder state; it also merges the two traits into a single
`Codec` trait, since they are currently always used together.

### DIFF
--- a/src/io/frame.rs
+++ b/src/io/frame.rs
@@ -1,5 +1,4 @@
 use std::io;
-use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 
@@ -185,17 +184,24 @@ impl<'a> Drop for EasyBufMut<'a> {
     }
 }
 
-/// Decoding of a frame from an internal buffer.
+/// Encoding and decoding of frames via buffers.
 ///
-/// This trait is used when constructing an instance of `Framed`. It defines how
-/// to decode the incoming bytes on a stream to the specified type of frame for
-/// that framed I/O stream.
+/// This trait is used when constructing an instance of `Framed`. It provides
+/// two types: `In`, for decoded input frames, and `Out`, for outgoing frames
+/// that need to be encoded. It also provides methods to actually perform the
+/// encoding and decoding, which work with corresponding buffer types.
 ///
-/// The primary method of this trait, `decode`, attempts to decode a
-/// frame from a buffer of bytes. It has the option of returning `NotReady`,
-/// indicating that more bytes need to be read before decoding can
-/// continue.
-pub trait Decode: Sized {
+/// The trait itself is implemented on a type that can track state for decoding
+/// or encoding, which is particularly useful for streaming parsers. In many
+/// cases, though, this type will simply be a unit struct (e.g. `struct
+/// HttpCodec`).
+pub trait Codec {
+    /// The type of decoded frames.
+    type In;
+
+    /// The type of frames to be encoded.
+    type Out;
+
     /// Attempts to decode a frame from the provided buffer of bytes.
     ///
     /// This method is called by `Framed` whenever bytes are ready to be parsed.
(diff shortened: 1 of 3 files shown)


aturon:
cc @tikue

tailhook:
It's unclear from the text what was the motivation to put both Encode+Decode to a single value? In FramedIo they were separate.

Not that I'm against that decision, but I'm not sure..

aturon:
For the `framed` helper, it seemed simpler and didn't seem to have a lot of downside (since if there's state on either side, you should be able to package it up). I was also wondering whether the traits themselves should be merged into a single `Codec` trait, but I suspect they'll end up useful broken up.

@tailhook Any thoughts/preferences?

tailhook:
I can't remember at the top of my head any protocols which can have different/interchangeable protocols for input and output.

On the other hand, I can remember a few protocols which are half-duplex. So separate traits should be useful.

aturon:
Oh, to be clear, a merged `Codec` trait would still have separate types -- it'd be totally equivalent to `Encode + Decode`, but mildly easier to implement/bound by.

tailhook:
Well, now I'm not sure what you mean :)

`trait Codec: Encode + Decode {}` ? I'm not sure it might be too useful.

On the other hand, if you merge the two traits into one then there are no good way to implement unidirectional protocol. Use `Void` as one of the types?

(If it wasn't clear, I'm in favor of the current version of the pull request in both aspects)

aturon:
@tailhook Hah, sorry for all the confusion. I'm obviously also in favor of the current PR, so let's leave it at that for now :-)

tikue:
LGTM!

alexcrichton:
Looks good to me! I'd personally be in favor of just fusing `Encode` and `Decode` to `Codec` already as they're not buying much as separate traits now anyway.

aturon:
Updated to merge the traits into a single `Codec` trait, given that they are always used together. If we end up finding uses within the stack for split traits, we can explore that at some later point.

src/io/frame.rs
```diff
@@ -216,57 +222,94 @@ pub trait Decode: Sized {
...
-        ReadState {
-            eof: false,
-            is_readable: false,
-            rd: EasyBuf::new(),
+impl<T: Io, C: Codec> Stream for FramedRead<T, C> {
+    type Item = C::In;
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<Option<C::In>, io::Error> {
+        if let Async::Ready(mut guard) = self.framed.poll_lock() {
```
alexcrichton, line 260: This is where I _slightly_ regret `if let`, the `match` clause I feel is more concise here:

``` rust
match self.framed.poll_lock() {
    Async::Ready(mut guard) => guard.poll(),
    Async::NotReady => Ok(Async::NotReady),
}
```

Just interesting trivia!

alexcrichton (commented):
👍

tailhook:
Okay, let me ask again, what about unidirectional protocols? There are a couple of them, for example protocol of [feeding metrics to carbon](http://graphite.readthedocs.io/en/latest/feeding-carbon.html).

While there is only `framed()` method in this PR, it's easy to make `framed_unidirectional_input()` or similar. When using combined codec protocol you need `Out=Void` and implement decode that always returns `None` which is ugly.

alexcrichton:
@tailhook in some sense this isn't necessarily intended to be the "end all be all" of how you frame data coming of or going onto the wire. In that sense if unidirectional protocols are quite common then it may make sense to split the trait again and add methods that only work with each half, but for now I'd imagine that those kinds of protocols are much rarer? In that sense it's certainly possible with the stream/sink APIs given by `framed` if not the most ergonomic.

tailhook:
Yes, they are much rarer. On the other hand, I don't see too much cost in using `Decode + Encode` comparing to `Codec`. 

Anyway, if you think that use case may be ignored I'm not going to argue to death.

aturon:
@tailhook Yep, point taken -- I think the goal right now is to provide a smooth path for the "mainstream" cases and try to get an overall 0.1 release out the door.

liigo:
The name of `In` and `Out` are a bit confusing.

``` rust
// In is used as output/return value
fn decode(&mut self, buf: &mut EasyBuf) -> Result<Option<Self::In>, io::Error>

// Out is used as an input parameter
fn encode(&mut self, msg: Self::Out, buf: &mut Vec<u8>) 
```

PR #104, opened 2016-11-16
Clean up Sink implementation on Framed

danburkert:
This commit makes a few changes to the Sink implementation on Framed:

* Backpressure is implemented for `start_send`. If the write buffer is
  over 8KiB and can't be flushed, no new items are accepted.
* 0 length writes to the upstream transport are translated into a
  `WriteZero` error, as with `io::Write::write_all`. `write_all` checks
  for and ignores `Interrupted` errors, but I do not think this is
  necessary for non-blocking writes.
* Potentially fewer calls to prefix truncate the write buffer in
  `poll_complete`.
* In `poll_complete`, the upstream transport is not flushed until
  *after* writing the entire write buffer.

### DIFF
--- a/src/io/frame.rs
+++ b/src/io/frame.rs
@@ -357,27 +357,37 @@ impl<T: Io, C: Codec> Sink for Framed<T, C> {
     type SinkError = io::Error;
 
     fn start_send(&mut self, item: C::Out) -> StartSend<C::Out, io::Error> {
+        // If the buffer is already over 8KiB, then attempt to flush it. If after flushing it's
+        // *still* over 8KiB, then apply backpressure (reject the send).
+        if self.wr.len() > 8 * 1024 {
+            try!(self.poll_complete());
+            if self.wr.len() > 8 * 1024 {
+                return Ok(AsyncSink::NotReady(item));
+            }
+        }
+
         self.codec.encode(item, &mut self.wr);
         Ok(AsyncSink::Ready)
     }
 
     fn poll_complete(&mut self) -> Poll<(), io::Error> {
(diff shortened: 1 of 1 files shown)


src/io/frame.rs
```diff
@@ -357,27 +357,49 @@ impl<T: Io, C: Codec> Sink for Framed<T, C> {
...
+                    self.wr.drain(..offset);
+                    if error.kind() == io::ErrorKind::WouldBlock {
+                        return Ok(Async::NotReady);
+                    } else {
+                        return Err(error);
+                    }
+                }
             }
+        }
+        self.wr.drain(..offset);
```
alexcrichton: The major difference between this loop and the previous loop seems to be deferring the call to `drain` until the very end, which seems to be an optimization? Initial testing I've done ended up never showing that improvement in profiles, so I'm curious if you were able to get speedups with a change like this?

alexcrichton:
Looks good to me, thanks!

src/io/frame.rs
danburkert: I haven't tested, but I  would only expect it to show up if the upstream channel has a small buffer (unlike Tcp).  The main thing here was fixing the lack of backpressure and potential infinite loop on write zero - I can simplify this back to the way it was if you want.

alexcrichton: Ok let's stick to that for now. I'm totally fine optimizing this loop but I'd just prefer to have some numbers to see where it matters first.

danburkert:
Removed the drain coalescing.

alexcrichton:
Thanks!

PR #106, opened 2016-11-16
Implement `TcpListener::accept()`

dpc:
Would something like this be accepted? I'm aware it's not an idiomatic way, but sometimes there's a need to just attempt to accept one connection. My need is to support fibers/coroutines as `Futures`, where a fiber is accepting connections one by one: https://github.com/dpc/tokio-fiber/blob/aef185bac414fb1745e2c1addf813fae981dc71e/examples/echo2.rs#L51

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -145,38 +145,17 @@ impl TcpListener {
         }
 
         impl Stream for MyIncoming {
-            type Item = (mio::tcp::TcpStream, SocketAddr);
+            type Item = (TcpStream, SocketAddr);
             type Error = io::Error;
 
             fn poll(&mut self) -> Poll<Option<Self::Item>, io::Error> {
-                if let Async::NotReady = self.inner.io.poll_read() {
-                    return Ok(Async::NotReady)
-                }
-                match self.inner.io.get_ref().accept() {
-                    Ok(pair) => Ok(Async::Ready(Some(pair))),
-                    Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
-                        self.inner.io.need_read();
-                        Ok(Async::NotReady)
-                    }
-                    Err(e) => Err(e)
-                }
+                Ok(Async::Ready(Some(try_nb!(self.inner.accept()))))
             }
         }
 
-        let remote = self.io.remote().clone();
         let stream = MyIncoming { inner: self };
         Incoming {
-            inner: stream.and_then(move |(tcp, addr)| {
-                let (tx, rx) = futures::oneshot();
-                remote.spawn(move |handle| {
-                    let res = PollEvented::new(tcp, handle).map(move |io| {
-                        (TcpStream { io: io }, addr)
-                    });
-                    tx.complete(res);
-                    Ok(())
-                });
-                rx.then(|r| r.expect("shouldn't be canceled"))
-            }).boxed(),
+            inner: stream.boxed(),
         }
     }
 
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -34,6 +34,39 @@ impl TcpListener {
         TcpListener::new(l, handle)
     }
 
+    /// Attempt to accept a connection and create a new connected `TcpStream` if successful.
+    ///
+    /// It is more idiomatic to treat incoming connection as a `Stream` of `TcpStream`s.
+    /// See `incoming()` for details.
+    pub fn accept(&self) -> io::Result<(TcpStream, SocketAddr)> {
+        if let Async::NotReady = self.io.poll_read() {
+            return Err(io::Error::new(io::ErrorKind::WouldBlock, "not ready"))
+        }
+
+        let res = self.io.get_ref().accept();
+        match res {
+            Err(e) => {
+                if e.kind() == io::ErrorKind::WouldBlock {
+                    self.io.need_read();
+                }
+                Err(e)
+            },
+            Ok((sock, addr)) => {
+                let (tx, rx) = futures::oneshot();
+                let remote = self.io.remote().clone();
+                remote.spawn(move |handle| {
+                    let res = PollEvented::new(sock, handle)
+                        .map(move |io| {
+                            (TcpStream { io: io }, addr)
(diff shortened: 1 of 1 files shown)


alexcrichton:
I'd be totally down for having this! Perhaps through a `Handle` could be passed in to avoid the dance with `oneshot`, `spawn`, and `wait`?

dpc:
TBH, I'm not sure if I fully understand the whole `Handle` & `Remote` thing. From my (user) perspective it might be an inconvenience, as fibers / coroutines are considered thread-like, so things inside them must be passed as `Send`. And as `Handle` is not `Send`, so one have to use a `Remote` and then do similar "dance" manually. So if I understand it right, I would have to change: https://github.com/dpc/tokio-fiber/blob/aef185bac414fb1745e2c1addf813fae981dc71e/examples/echo2.rs#L51 line  to include such "dance"

Similar problem with `TcpListener::bind`. I guess some wrapper that do the dance and returns the value back could be created. Or two invocations provided - one taking `Handler`, one that doesn't need it and do the "dance" .

alexcrichton:
A `Handle` is basically just a standard handle to an event loop. With one you can create I/O objects or otherwise send messages to the event loop. The critical difference, however, is that `Handle` cannot be sent across threads whereas a `Remote` can be. Because a `Remote` can be sent across threads it has less functionality than a `Handle` unfortunately, and primarily only allows spawning futures rather than creating I/O objects directly.

The problem with the current construction is that in theory `accept` is nonblocking but it is in fact _blocking_. If that code path is executed on the event loop itself it risks blocking the I/O thread. 

The solution would be to pass in a `Handle` directly to define which event loop the accepted connection should be accepted with. It's a bit wonky, yeah, unfortunately. I wonder, presumably you'd want easy interoperation between fibers and futures, right? That is, while executing a fiber, if handed a future it should know how to wait for the future in a "cooperative" fashion?

dpc:
I wasn't aware that `accept` can block. I expected it to just resturn `WouldBlock` if there's nothing to accept.

Yes. The goal is to do exactly what the https://github.com/dpc/tokio-fiber/blob/aef185bac414fb1745e2c1addf813fae981dc71e/examples/echo2.rs#L51 is trying to do - have a code that is just like a blocking, threaded code, that just cooperatively jumps out of coroutine, when its blocked.

alexcrichton:
Oh well mio's `accept` method doesn't block. The one you've written here blocks because it calls `Future::wait`.

If the future interop is possible then I see a few possible solutions here:
1. With the future interop enable stream interop, enabling usage of `incoming`.
2. Take a `&Handle` argument to associate accepted connections.
3. Modify this method to track the internal state of the spawned future. That is, keep the impl roughly the same as it is today, it just calls `poll` on the spawned future in `accept` sometimes if one was previously scheduled.

dpc:
Ah, I see what you mean now. I was considering that `wait` harmless, but now I see it isn't.
1. Is definitely on my list.

However what if `accept` returned a future resolving to `TcpStream`? Do you think it would work OK with other usage? Since `tokio-fiber` already has `await`, the coroutine code would change from:

```
 let (mut conn, _addr) = poll!(socket.accept())?;
```

to

```
 let (mut conn, _addr) = await(socket.accept()))?;
```

alexcrichton:
Yeah `accept` could return a future for sure. That may be the best option in the end. I was hoping to keep the I/O operations all pretty consistent amongst one another in the sense that currently they all just operate in the standard fashion (return an `io::Result`) where an error is returned on "would block". If we were to return a future from `accept` it'd be a deviation from that as that method would return a future rather than an `io::Result`.

I will say though this looks pretty slick:

``` rust
let (mut conn, _addr) = await(socket.accept()))?;
```

dpc:
There can be multiple versions (with handle, returning socket, and another not taking handle but returning future). Let me know what version would you be willing to merge ATM (it can be changed later too, I guess) and I'll modify the PR accordingly.

alexcrichton:
Hm so thinking some more, I don't think this should return a future because then we have to deal with the semantics of what happens if you call it twice in a row, and in general that's what the `incoming` stream is for anyway.

Additionally I don't think this should take a `Handle` as that would be inconsistent.

Let's see if we can have this signature (as is in this PR) but change the implementation to keep ahold of the spawned future and poll that on future calls to `accept`.

Does that sound ok?

dpc:
Sure. I will give it a try.

dpc:
I had to make it `&mut self`. I think it's consistent with other IO operations (`read` and `write`). Alternatively `RefCell` could be used.

alexcrichton:
Looks great to me! Could the `incoming` method also be defined in terms of this method calling `accept`?

dpc:
Oh. Haven't thought about it (plus I was too shy to touch existing code). Will check.

src/net/tcp.rs
```diff
@@ -98,38 +145,17 @@ impl TcpListener {
...
-                remote.spawn(move |handle| {
-                    let res = PollEvented::new(tcp, handle).map(move |io| {
-                        (TcpStream { io: io }, addr)
-                    });
-                    tx.complete(res);
-                    Ok(())
-                });
-                rx.then(|r| r.expect("shouldn't be canceled"))
-            }).boxed(),
+            inner: stream.boxed(),
```
dpc, line 158: It seems to me that boxing here is unnecessary if `MyIncoming` is moved outside of the function or merged with `Incoming` altogether. Can do in a separate patch.

alexcrichton:
Oh sorry I missed the update here! Looks good to me, so merging.

alexcrichton:
Er nvmd this was literally just updated as I looked at it, what a coincidence!

dpc:
Yeah. You're fast. :)

src/net/tcp.rs
alexcrichton: Ah yeah we can postpone this to a future PR

dpc:
Should I bother with removing that one `boxed`?

alexcrichton:
Ah it's ok it was easy enough to just push up locally, so I've already done so now.

PR #109, opened 2016-11-20
Udp frame

rrichardson:
This implements FramedUdp for UdpSocket, which mirrors the Framed module for connection based protocols.  This is a continuation of #35


rrichardson:
OK. Everything is implemented, but I'm unsure about the Bilock for TCP and UDP Stream/Sink. It seems to me that in most cases, sending and receiving simultaneously on most all Fds is perfectly OK. If its not, the OS typically brokers the connection, failing that, we can use the Bilock on specific implementations where we know it is necessary.

The current implementation could lead to slowdowns, where we're needlessly limiting parallelism. In addition, it seems like its the wrong abstraction, when what we need is to ensure that multiple copies of Stream or Sink are not used simultaneously. 

I reckon there might still be additional hurdles re `Send`, and there is obviously the danger that a user could close the fd in one thread, while sending in another, but that would kind of mean that we should move refcounting to the fd that would protect it in the case of against `close()` and `drop()`

I'm tempted to just clone the socket and leave out the Bilock for FramedUdp. We could even create separate Arcs for each of the Stream and Sink.

Thoughts?

src/io/udp_frame.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// `Ok(None)` is returned. This indicates to the `Framed` instance that
+    /// it needs to read some more bytes before calling this method again.
+    /// In such a case, it is decode's responsibility to copy the data
+    /// into their own internal buffer for future use.
+    ///
+    /// Finally, if the bytes in the buffer are malformed then an error is
+    /// returned indicating why. This informs `Framed` that the stream is now
+    /// corrupt and should be terminated.
+    ///
+    fn decode(&mut self, src: &SocketAddr, buf: &[u8]) -> Result<Option<Self::In>, io::Error>;
```
alexcrichton: I'm a little worried about this returning an `Option` because I'm not sure what the implementation of `Framed` can do if it returns `None`. Right now it just appends more data to the internal vector but the original `SocketAddr` is lost, meaning that this can be called with data from two different addresses (which I don't think is desired).

I was thinking this would return `Self::In` instead of an `Option` to signify that it always needed to succeed. The codec could perhaps handle partial frames by buffering internally, though?

src/io/udp_frame.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+impl<C : CodecUdp> Stream for FramedUdp<C> {
+    type Item = C::In;
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<Option<C::In>, io::Error> {
+        loop {
+
+            let ret = self.socket.recv_from(self.rd.as_mut_slice());
```
alexcrichton, line 74: This can probably be simplified with:

``` rust
let (n, addr) = try_ready!(self.socket.recv_from(...));
```

src/io/udp_frame.rs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        self.out_addr = Some(self.codec.encode(&item, &mut self.wr));
+        Ok(AsyncSink::Ready)
+    }
+
+    fn poll_complete(&mut self) -> Poll<(), io::Error> {
+        trace!("flushing framed transport");
+
+        while !self.wr.is_empty() {
```
alexcrichton: This could just be an `if`, right? (rather than a loop)

src/io/udp_frame.rs
```diff
@@ -0,0 +1,251 @@
...
+                self.wr.clear();
+                self.out_addr = None;
+                if n != remaining {
+                    return Err(io::Error::new(io::ErrorKind::WriteZero,
+                                              "failed to write frame datagram to socket"));
+                }
+            }
+            else {
+                return Err(io::Error::new(io::ErrorKind::Other,
+                                          "outbound stream in invalid state: out_addr is not known"));
```
alexcrichton: I think this should be a panic?

src/io/udp_frame.rs
```diff
@@ -0,0 +1,251 @@
...
+        while !self.wr.is_empty() {
+            if let Some(outaddr) = self.out_addr {
+                let remaining = self.wr.len();
+                trace!("writing; remaining={}", remaining);
+                let n = try_nb!(self.socket.send_to(self.wr.as_slice(), &outaddr));
+                trace!("written {}", n);
+                self.wr.clear();
+                self.out_addr = None;
+                if n != remaining {
+                    return Err(io::Error::new(io::ErrorKind::WriteZero,
```
alexcrichton: I'd be fine with just an `Other` error kind here for now as `WriteZero` isn't _quite_ what happened.

src/io/udp_frame.rs
```diff
@@ -0,0 +1,251 @@
...
+        vec![0; 64 * 1024],
+        Vec::with_capacity(64 * 1024)
+    )
+}
+
+impl<C> FramedUdp<C> {
+
+    /// Creates a new FramedUdp object.  It moves the supplied socket, codec
+    /// supplied vecs. 
+    pub fn new(sock : UdpSocket, codec : C, rd_buf : Vec<u8>, wr_buf : Vec<u8>) -> FramedUdp<C> {
```
alexcrichton: Let's keep this private for now and just expose the `framed` constructor on `UdpSocket` directly.

src/net/udp.rs
```diff
@@ -85,6 +98,33 @@ impl UdpSocket {
...
+    /// needed).
+    ///
+    /// Any error which happens during writing will cause both the stream and the
+    /// buffer to get destroyed.
+    ///
+    /// The `buf` parameter here only requires the `AsRef<[u8]>` trait, which should
+    /// be broadly applicable to accepting data which can be converted to a slice.
+    /// The `Window` struct is also available in this crate to provide a different
+    /// window into a slice if necessary.
+    pub fn send_dgram<'a, T>(&'a self, buf: T, addr : &'a SocketAddr) -> SendDGram<T>
```
alexcrichton: For consistency with other helpers in `tokio_core::io` perhaps this could take self by value and not contain an internal borrow?

src/net/udp.rs
```diff
@@ -249,6 +289,61 @@ impl fmt::Debug for UdpSocket {
...
+    type Item = T;
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<T, io::Error> {
+        match self.state {
+            UdpState::Writing { ref sock, ref buf, ref addr} => {
+                let buf = buf.as_ref();
+                let n = try_nb!(sock.send_to(&buf, addr));
+                if n == 0 {
+                    return Err(zero_write())
```
alexcrichton: It's probably fine to just have the clause below rather than singling out the zero case

src/io/udp_frame.rs
rrichardson: I think the receive -> decode behavior in datagrams should be pretty much the same for streams. 

If, in the case that some protocol frame is divided up into N chunks, when we receive chunk < N,
it seems like it would be good to indicate that we did receive data, and we are accumulating it in the codec until we can return a complete frame.  

However, in either case, it looks like the behavior to the consumer (Stream) would be the same, which is
`return Ok(Async::NotReady)`

The implementor of the codec would be responsible to associate the incomplete message with the inbound/outbound socketaddr.  

Were you thinking that the handling of chunked messages should be done by the user of FramedUdp rather than the implementor of Codec?

src/io/udp_frame.rs
rrichardson: <del>yep. much simpler. </del> 

Actually no, recv_from follows the same semantics as tcp::read, which returns an Io::Result

src/io/udp_frame.rs
rrichardson: I wasn't sure about this one,  since   UdpSocket::framed requires a circular dependency betwen  net/udp.rs  and io/udp_framed.rs.

I was taught that when someone asks you to create circular dependencies between API layers, you should run away screaming loudly, and tell somebody you trust.

src/net/udp.rs
rrichardson: Agreed.

src/io/udp_frame.rs
rrichardson: Fixed.

src/io/udp_frame.rs
rrichardson: Fixed.

src/io/udp_frame.rs
rrichardson: SGTM

src/net/udp.rs
rrichardson: Done.

src/io/udp_frame.rs
rrichardson: I'll go ahead and make it private.. we can worry about API layers later I guess.

rrichardson:
Ok.  I made pretty much all suggested updates except for removing Option from the result of decode, I can still be convinced to do it. I'm just awaiting more feedback.

alexcrichton:
So the streaming case is definitely interesting. So let's say you've got a protocol with messages both coming in and going out which are larger than the datagram size. This means that on both ends you'll have to split it up.

In that sense it makes sense for `decode` to return an `Option`, maybe buffer internally. Conversely, though, `encode` needs to be able to send multiple datagrams. Right  now it seems like it may be hard for `encode` to send multiple datagrams though? (as I don't think we'd want to automatically slice it up).

That leads me to the conclusion that we should stick with taking `Self::Out` and returning `io::Result<Self::In>`. That basically means:

* Every datagram we read must be a complete "frame"
* Every datagram we write must also be a complete frame.

If a protocol then implements a form of streaming it'd do that by parsing out individual frames and manage the reassembly and disassembly itself. I think something like this is where tokio-proto would come in. (you'd presumably also want to handle cases where you receive one part of a stream but then timeout out receiving the rest, which seems hard to do at this layer).

Does that make sense for removing the `Option` for now? To handle streaming at a different layer?

---

On and I missed the part about `BiLock` and how it's not necessary for UDP. You're 100% right that it's not necessary in all cases. TCP and/or UDP isn't always the lowest layer though where a `BiLock` is definitely necessary. We figured it's best to punt on this performance for now as it seems like the overhead would be negligible. If you see it show up though we'd be quite interested in the use case!

Note that `Stream::split` was added recently, so we can likely delete the `split` method here in favor of that. Eventually we could specialize `split` to avoid a `BiLock` and use an `Arc` internally perhaps with specialization even if we find the right impls available.

rrichardson:
Seems reasonable, I reckon.   I could definitely see a case where encode would need to return more than one frame/datagram if we actually supported the chunking/streaming model. 

In addition, a professional installation of a protocol would probably discover the Path MTU for a host and adjust the chunk size accordingly.  This would make for a complicated codec, as it would need access to things it doesn't over which it doesn't have purview. 

I guess this means I will also supply a default codec which produces Vec<u8> of the complete datagram, as there really isn't any reason to do anything different unless someone wants custom buffers for their datagrams.

alexcrichton:
Ok! I'll merge and I think I'll clean up a thing or two along the way. Thanks again for the PR!

Issue #108, opened 2016-11-20
Codec "encode" return value


colindjk:
Currently the encode function takes a buffer to write to as a parameter.

I was wondering if it would be possible to have the encode function return a result just as it did previously, prior to the Encode and Decode getting pushed into a unified Codec trait.

Unless of course there is justification, but I really wouldn't see one as the benefits of being able to notify the user of a failed "encode" would be real important.

alexcrichton:
Hm I'm not sure that encode was able to ever signal an error, but it seems like an ok ability to me! Could you elaborate on the use case as well? Or do you think it should just return an `io::Result<()>`?

colindjk:
Well yeah I was just thinking maybe an invalid character was given.

What I have now is a situation where I'm encoding a TCPStream into frames. The 'frames' are just JSON objects, so I encode the object into the said frame (JSON object). 

So there's the issue actually, in order to get that serialized info, I need to copy it into the buffer given by encode! (Yuck). That means serializing into a buffer, then copying that buffer data into a given buffer. To top it off, in order to serialize, I need to handle the possible errors inherent with the serialize functions.

Unfortunately though, the only way I can handle errors is by returning an empty array. Which then requires me to check the return value of encode so that the length of the array is not in fact zero.

Well jeepers, even writing that feels un-rust XD.

Anyway if it could just return io::Result<()> would be enough for me at least. There just needs to be something to pass up the error from whatever encoder I decide to use... I could make a pull request for ya :3.

colindjk:
Oops ignore that first sentence it's not supposed to be there....

PR #110, opened 2016-11-21
Updated 'encode' return value to Result, removed parameter passing bu…

colindjk:
Issue #108, there should be a return value for two reasons
    - Avoid copying arrays from serialize functions
    - Allow for error handling for unique encode functions.

### DIFF
--- a/src/io/frame.rs
+++ b/src/io/frame.rs
@@ -243,7 +243,7 @@ pub trait Codec {
     /// This method will encode `msg` into the byte buffer provided by `buf`.
     /// The `buf` provided is an internal buffer of the `Framed` instance and
     /// will be written out when possible.
-    fn encode(&mut self, msg: Self::Out, buf: &mut Vec<u8>);
+    fn encode(&mut self, msg: Self::Out) -> Result<Vec<u8>, io::Error>;
 }
 
 /// A `Stream` interface to an underlying `Io` object, using the `Decode` trait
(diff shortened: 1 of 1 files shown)


src/io/frame.rs
```diff
@@ -243,7 +243,7 @@ pub trait Codec {
     /// This method will encode `msg` into the byte buffer provided by `buf`.
     /// The `buf` provided is an internal buffer of the `Framed` instance and
     /// will be written out when possible.
-    fn encode(&mut self, msg: Self::Out, buf: &mut Vec<u8>);
+    fn encode(&mut self, msg: Self::Out) -> Result<Vec<u8>, io::Error>;
```
alexcrichton, line 246: Here it's actually important to take a destination buffer as an argument to avoid double-buffering or copying data unnecessarily. Perhaps this could still take `&mut Vec<u8>` but it could return `io::Result<()>`?

src/io/frame.rs
```diff
@@ -366,7 +366,7 @@ impl<T: Io, C: Codec> Sink for Framed<T, C> {
             }
         }
 
-        self.codec.encode(item, &mut self.wr);
+        self.wr = self.codec.encode(item).unwrap();
```
alexcrichton, line 369: Here I believe we'll want to return the error to the caller rather than calling `unwrap`

PR #114, opened 2016-11-25
add ToSocketAddrs on TcpStream

danylaporte:
Use the same pattern as in the std lib for creating a TcpStream.
I am new to rust, please make sure it is correct and idiomatic to do so.


alexcrichton:
Thanks for the PR! Unfortunately though `ToSocketAddrs` isn't quite appropriately for tokio-core right now. That trait in the standard library may do synchronous DNS resolution which can unnecessarily block the reactor. We're looking to add this functionality one day through an asynchronous DNS resolver trait, but we aren't quite there yet, unfortunately.

PR #115, opened 2016-11-26
Generalize datagram operations (send_to, recv_from) in order to enable Futures for other datagram socket types

dermesser:
In your library `tokio-uds`, the [`UnixDatagram`](https://tokio-rs.github.io/tokio-uds/tokio_uds/struct.UnixDatagram.html) type lacks an interface allowing easy combination with futures (users need to write their own glue code).

By generalizing `send_to()` and `recv_from()` into separate traits, crates like `tokio-uds` could use the code that is already present in `SendDgram` and `RecvDgram` in order to enable an interface working with futures.

My current plan is to later submit a patch to `tokio-uds` that makes use of what I'm changing here. I'm also not entirely certain about the location of the dgram module -- should it stay in `net/udp`? 

Another downside is that existing users of `recv_from()`/`send_to()` would need to adapt their imports if they want to upgrade.

Please let me know what you think; I'm open to suggestions about how to improve this!


alexcrichton:
Thanks for the PR! I initially shied away from traits here, however, as I don't think they'll end up being particularly useful. The primary point of contention, `SocketAddr`, isn't shared amongst UDP and UDS. In that sense, I don't think these traits can be implemented for unix sockets?

dermesser:
ah, you're absolutely right. For some reason I had in mind that SocketAddr included unix addresses (maybe because the syscalls [work that way](http://man7.org/linux/man-pages/man2/connect.2.html)).

Anyway, nevermind then :-)

dermesser:
On a related note -- would you still accept a PR implementing recv_dgram()/send_dgram() methods for UDS in tokio-uds?

alexcrichton:
Certainly!

PR #116, opened 2016-12-02
Add example demonstrating the use of the RecvDgram future

dermesser:
The exact usage is a bit non-trivial, and this example may be helpful in understanding. Let me know if you have suggestions.


alexcrichton:
Nice idea! I wonder if we could perhaps have a slightly different example though because this seems actually more appropriate as `framed` (processing as a stream). The combinators are intended for one-off use cases, so perhaps a simple ping/pong server running?

dermesser:
hmm, I think I just misunderstood that part of the API. It's probably not necessary; and I can't come up with a nice example for a one-off ping/pong server anyway :-\

PR #118, opened 2016-12-06
Handle `hup` events as indicating read readiness

rotty:
Using Linux's `epoll(7)` interface, a `EPOLLHUP` condition is
signalled for the reading end of a pipe or socket when the other end
is closed for writing. This may happen in combination with `EPOLLIN`
being signalled (if further data is available for reading), or without
`EPOLLIN`.

If `EPOLLHUP` is signalled without `EPOLLIN` it indicates an immediate
EOF condition, which will result in the next `read()` suceeding with 0
bytes read. It is thus not required to handle `EPOLLHUP` specially in
the reader, but we need to indicate readiness upon encountering it.

Looking at the `kqueue` and `windows` mio backends, they seem to be
turn a detected EOF and connection reset into `mio::Ready::hup()`
events, so it may be reasonable to speculate that this change is an
improvement for these platforms as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,3 +20,4 @@ slab = "0.3"
 
 [dev-dependencies]
 env_logger = { version = "0.3", default-features = false }
+nix = "0.7"
--- a/src/reactor/mod.rs
+++ b/src/reactor/mod.rs
@@ -306,7 +306,7 @@ impl Core {
         let mut writer = None;
         let mut inner = self.inner.borrow_mut();
         if let Some(io) = inner.io_dispatch.get_mut(token) {
-            if ready.is_readable() {
+            if ready.is_readable() || ready.is_hup() {
                 reader = io.reader.take();
                 io.readiness.fetch_or(1, Ordering::Relaxed);
             }
@@ -430,7 +430,7 @@ impl Inner {
         let entry = self.io_dispatch.vacant_entry().unwrap();
         try!(self.io.register(source,
                               mio::Token(TOKEN_START + entry.index() * 2),
-                              mio::Ready::readable() | mio::Ready::writable(),
+                              mio::Ready::readable() | mio::Ready::writable() | mio::Ready::hup(),
                               mio::PollOpt::edge()));
         Ok((sched.readiness.clone(), entry.insert(sched).index()))
     }
--- a/tests/pipe-hup.rs
+++ b/tests/pipe-hup.rs
@@ -1,0 +1,115 @@
+#![cfg(unix)]
+
+extern crate env_logger;
+extern crate futures;
+extern crate nix;
+extern crate mio;
+extern crate tokio_core;
+
+use std::io;
+use std::os::unix::io::{AsRawFd, RawFd};
+use std::time::Duration;
+use std::thread;
+
+use mio::{Evented, PollOpt, Ready, Token};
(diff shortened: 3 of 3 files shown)


alexcrichton:
Thanks! Would it be possible to also add a test for this?

rotty:
I'll try to come up something inspired by the tokio-process code.

rotty:
It seems that `epoll()` signalling just `EPOLLHUP`, without including `EPOLLIN` for that FD as well, can be only triggered with pipes. I have now have a test that excercises that case, but it duplicates quite some code from alexcrichton/tokio-process#1. In particular, the duplicated code provides the machinery to wrap a Unix `RawFd` in a type implementing `PollEvented`. 

This seems like something that makes sense to expose as Unix-specific API in tokio-core, but I'm not sure whether I'm going in the right direction with that idea, and probably would need some guidance on the API surface design, should I prepare a PR for that feature. So, if this is deemed a reasonable idea, I can prepare a rough pull request, and the API bikeshedding can start ;-). @alexcrichton Any ideas on that topic?

rotty:
Note that I've now included a rough sketch of the test as a separate, preceding commit so it's clear what kind of code I'm talking about.

Cargo.toml
```diff
@@ -12,11 +12,15 @@ the rest of the tokio crates.
...
 mio = "0.6"
 scoped-tls = "0.1.0"
 slab = "0.3"
 
 [dev-dependencies]
 env_logger = { version = "0.3", default-features = false }
+nix = "0.7"
+
+[replace]
+'futures:0.1.3' = { git = "https://github.com/alexcrichton/futures-rs" }
```
alexcrichton: I think these changes may be causing CI to fail?

Cargo.toml
```diff
@@ -12,11 +12,15 @@ the rest of the tokio crates.
...
-futures = "0.1.4"
+futures = "0.1"
 log = "0.3"
 mio = "0.6"
 scoped-tls = "0.1.0"
 slab = "0.3"
 
 [dev-dependencies]
 env_logger = { version = "0.3", default-features = false }
+nix = "0.7"
```
alexcrichton, line 23: For now could `libc` be used instead?

alexcrichton (commented):
Test looks good to me! I think you may also want to gate it with `#![cfg(unix)]` though to pass on Windows.

Cargo.toml
rotty: Hmm, it seems libc lacks `errno`. Bringing that into the test, in addition to all the boilerplate already there (for wrapping `RawFd`) would be over the top, at least for my taste. Also, the test would further lose readability, as the `libc` interfaces seem quite "raw" compared to their `nix` wrappers.

rotty:
Besides the `nix` vs. `libc` issue, I consider this fit for merging now.

alexcrichton:
I touched up the test a bit but otherwise looks great to me, thanks again!

Cargo.toml
rotty: I managed to not find `io::Error::last_os_error()`, so these "arguments" of mine mostly fall flat, as I see from your cleanup patch.

alexcrichton: Ah yeah no worries!

PR #125, opened 2016-12-16
add Core::with_capacity() method

### LINKED ISSUE #124: reactor::Core::new() always allocates ~20MB
Currently the initial slab capacity is [hardcoded to 64K](https://github.com/tokio-rs/tokio-core/blob/f036691681ae7868fc60480f99b9f2b62f1a63eb/src/reactor/mod.rs#L38). This causes a new `reactor::Core` to consume upwards of 20MB of heap memory, even if the program only ever uses it to serve a small number of connections.

It should be possible to specify the desired starting capacity of a `Core` upon construction.

dwrensha:
Fixes #124.


alexcrichton:
Thanks for the PR! I've thought about this in the past as well as it takes *milliseconds* to complete `Core::new` in debug mode which seems bad. Out of curiosity, do you think we even need to have an initial capacity at all? It seems like it should be possible to create everything with zero capacity initially and it grows dynamically.

Start may be a tad slower, but in theory not all that slower if we grow correctly.

dwrensha:
> Out of curiosity, do you think we even need to have an initial capacity at all?

Ah! That does sound like a better idea, especially because this `capacity` does not have a particularly precise meaning outside of the internal implementation details of `reactor::Core`. I've submitted #126, which I now prefer to the present pull request.

alexcrichton:
Ok, closing in favor of https://github.com/tokio-rs/tokio-core/pull/126

Thanks!

PR #126, opened 2016-12-16
initialize slabs with capacity of one

### LINKED ISSUE #124: reactor::Core::new() always allocates ~20MB

dwrensha:
Fixes #124. Note that the slabs cannot start with zero capacity, because then their dynamic capacity-doubling logic would fail.


alexcrichton:
Looks good to me, thanks!

Issue #133, opened 2016-12-21
Update TcpStream::new() and TcpStream::connect() to not use Box allocations


alreece45:
`TcpStream::new()` and `TcpStream::connect_stream()` use IoFuture-- which is a `Box`. They should probably be updated to return concrete types. 

Discussion on gitter suggests IoFuture may be removed in the future. https://github.com/alexcrichton/futures-rs/issues/228 suggests that `Futures::boxed()` may be removed as well.

This will probably be something for 0.2.0

alexcrichton:
This has been done!

PR #134, opened 2016-12-21
Unboxing TcpStream

alreece45:
Fairly rough work to unbox TcpStream's methods for #133 

Don't expect this to merge, but hope that it'll help someone who can clean it up a bit more-- especially some of the type-work figured out.

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -8,7 +8,7 @@ use futures::sync::oneshot;
 use futures::{self, Future, failed, Poll, Async};
 use mio;
 
-use io::{Io, IoFuture};
+use io::Io;
 use reactor::{Handle, PollEvented};
 
 /// An I/O object representing a TCP socket listening for incoming connections.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Perhaps we could just explicitly encode the state machine here with:

```rust
enum Connect {
    Connecting(PollEvented<TcpStream>)
    Error(io::Error),
    Empty,
}
```

and then `poll` would transition amongst those?

alreece45:
Have `Connect` be a public enum returned from `connect()` and `connect_stream()`?

That'd probably be simpler-- I resisted doing it because it changes the interface a bit more-- the enum would be public, more of the state would be exposed, and I don't have the expertise to know if that's a good decision in the long-term.

alexcrichton:
Oh I think we can keep the enum private, it'd just be a minor implementation detail of the `TcpStreamNew` return value. We'd want to keep the same API for now and if we want to break it we can consider that for 0.2

alexcrichton:
Ok I touched up a few of the commits here manually and merged, thanks again for the PR!

PR #139, opened 2017-01-04
Use `io::Result`

sinkuu:
`io::Result` is used instead of `result::Result<T, io::Error>` except this one.


pyfisch:
:+1: I also stumbled upon this.

alexcrichton:
Thanks!

PR #141, opened 2017-01-05
Update the use of deprecated APIs

critiqjo:
For stuff that moved from `futures::task` to `futures::executor`.

### DIFF
--- a/src/reactor/mod.rs
+++ b/src/reactor/mod.rs
@@ -14,7 +14,8 @@ use std::sync::atomic::{AtomicUsize, ATOMIC_USIZE_INIT, Ordering};
 use std::time::{Instant, Duration};
 
 use futures::{self, Future, IntoFuture, Async};
-use futures::task::{self, Unpark, Task, Spawn};
+use futures::executor::{self, Spawn, Unpark};
+use futures::task::Task;
 use mio;
 use slab::Slab;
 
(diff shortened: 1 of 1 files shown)


critiqjo:
Got a bit confused while reading through the code, and referring `futures` docs.

Aside from that, why is there `#![allow(deprecated)]` in [`src/channel.rs`](https://github.com/tokio-rs/tokio-core/blob/master/src/channel.rs#L7)? It compiles fine even if I make it `#![deny(deprecated)]`. Is it because the compiler, for some reason, ignores all deprecation flags outside `std`? (I did not get any warnings when compiling this crate for the deprecated usage this PR fixes.)

alexcrichton:
Thanks!

I'm not entirely sure how `#[deny(deprecated)]` works with the deprecated feature...

critiqjo:
I forgot to ask this at the time. The last paragraph in doc of `Core::run` contains [an outdated reference to `task::poll_on`](https://github.com/critiqjo/tokio-core/blob/e5bbf004d351f9829c192f792db7311ef88ab8ce/src/reactor/mod.rs#L205-L208). Is it possible to do that anymore? Since `poll` takes `&mut self`, how can one `poll` it from anywhere else? (I'm still in the process of learning the inner clockwork of futures. So please correct me if I am wrong.)

alexcrichton:
Ah right yeah that can likely just be entirely deleted, want to send a PR for that?

critiqjo:
> want to send a PR for that?

Done! #144

PR #152, opened 2017-01-20
Add from_raw_fd methods for TcpStream and TcpListener

GuillaumeGomez:
I use fd passing between processes and need to be able to "rebuild" the stream.


alexcrichton:
This looks good to me, but can you provide this functionality through extension traits rather than inherent methods?

GuillaumeGomez:
At the start, I wanted to add it through the `FromRaw` trait but the `Handle` was needed. How would want me to call the trait?

alexcrichton:
You'd create a new trait, like `TcpStreamExt`

GuillaumeGomez:
Strange name choice but let's go for `TcpStreamExt`! Also, don't you want to add `#![deny(warnings])` as well to tokio?

src/net/tcp.rs
```diff
@@ -492,6 +494,50 @@ impl Future for TcpStreamNewState {
...
+    unsafe fn from_raw_fd(fd: RawFd, handle: &Handle) -> Option<TcpStream> {
+        match PollEvented::new(mio::tcp::TcpStream::from_raw_fd(fd), handle) {
+            Ok(poll) => Some(TcpStream { io: poll }),
+            _ => None,
+        }
+    }
+}
+
+#[cfg(unix)]
+impl TcpStreamExt for TcpListener {
```
alexcrichton: This should be a `TcpListenerExt` trait.

Are you familiar with the extension trait pattern in the standard library? That's basically what's happening here.

alexcrichton: Also note that this isn't strictly necessary for TCP listeners due to the `TcpListener::from_listener` method, but I think it's fine to have it for consistency.

carllerche:
Would it be better to not have a conversion directly from RawFd (especially since it requires a new trait) and instead go Fd -> std::Tcp* -> tokio::Tcp*.

I don't currently see a `TcpStream::from_stream` fn, is there a reason that it doesn't exist? Anyway, I would rather see this strategy unless there is a good reason why it doesn't work.

GuillaumeGomez:
@carllerche: To create a tokio stream, you need a `Handle` and so creating a function (through a trait or not) anyway. So why not doing it at once?

src/net/tcp.rs
GuillaumeGomez: Seems more logical this way.

GuillaumeGomez:
@alexcrichton: Updated.

alexcrichton:
@carllerche that's a very good point yeah, and may sway me to preferring we add `from_stream` (no reason it doesn't currently exist) instead.

alexcrichton:
Note that another reason to not go the platform-specific extension route is that means the platform-specific extensions don't have to be duplicated in all the various crates. It may be more convenient ergonomically, but that's not really what we're optimizing for at this point.

GuillaumeGomez:
So... Do I need to update something?

alexcrichton:
I agree with @carllerche that this isn't the right strategy, so I'm going to close.

If you'd like to continue along this though then a PR which adds `TcpStream::from_stream` should do the trick.

Issue #120, opened 2016-12-08
Implement more traits for EasyBuf


Yamakaky:
`Clone`, `Deref`, `Into<Vec<u8>>` would be a start.

tailhook:
I think also `Index<..>` would be cool. Because `drain_to` is very-limited.

Yamakaky:
`Deref<[u8]>` would be sufficient, I think?

tailhook:
No, I mean slicing should return `EasyBuf` not `&[u8]`, because `EasyBuf` has a static lifetime, so can be a part of a return value of the protocol parser (and `EasyBuf` is specifically designed not to make a clone of the underlying data for that, it's how `drait_to` works)

Yamakaky:
Oh, yeah

alexcrichton:
The intention was to be *very* conservative for the initial iteration that we could tack on more functionality later, but I couldn't mind adding a few more common trait impls and such. I'd prefer to avoid `Deref` for now to avoid conflicts with the slice API, but other common traits should be good to go.

aturon:
cc https://github.com/tokio-rs/tokio-core/pull/128

manuel-woelker:
@tailhook I was looking at implementing Index for EasyBuf, with Output = EasyBuf. But it seems `index`
returns `&Self::output`. That means my naive approach - creating a new EasyBuf instance to return - won't work since it does not live long enough. Any pointers on how to proceed? Would this require `IndexGet`/`IndexMove`?

tailhook:
@manuel-woelker, hm, looks like we can't do that :(. The best way I can see is to have a method `x.slice(a..b)`.

alexcrichton:
Done in https://github.com/alexcrichton/tokio-io, the migration of this crate will be tracked by https://github.com/tokio-rs/tokio-core/issues/61

PR #154, opened 2017-01-22
impl fmt::Debug for EasyBuf (cf. #120)

manuel-woelker:
Debug format looks like this for small resp. long buffers

EasyBuf{len=2/6 [5, 6]}
EasyBuf{len=255/255 [0, 1, 2, 3, ..., 251, 252, 253, 254]}

I chose to include the size of the underlying vector buffer to easily detect situations where a small buffer is actually holding onto a large chunk of memory.

Feedback welcome!

### DIFF
--- a/src/io/frame.rs
+++ b/src/io/frame.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 use std::io;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks! Merged.

PR #155, opened 2017-01-23
Implement Deref for EasyBuf

mbrubeck:


alexcrichton:
I originally explicitly avoided this because I didn't want to conflate the API surface areas and wanted an explicit cast with `as_slice`, but could you elaborate more on the motivation for this?

mbrubeck:
My motivation was purely ergonomic, to avoid cluttering my code with explicit conversions.  It sounds like that's a non-goal, so feel free to close this.

alexcrichton:
Ok, in that case I'd prefer to game out the use cases over time to see if we still need this down the road. Thanks for the PR though!

PR #157, opened 2017-01-24
add read_bufs and write_bufs to Io and TcpStream

seanmonstar:
With this having landed in and released in mio, we can now take advantage of it on `TcpStream`. By adding it to the `Io` trait, programs that are generic over the Io object can still make use of it. The docs are partially lifted from the docs in mio.

The default implementation of `read_bufs` only uses the first buffer in the list, but it seems required to prevent the one call turning in to multiple syscalls.

### DIFF
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -14,6 +14,8 @@ use std::io;
 use futures::{BoxFuture, Async, Poll};
 use futures::stream::BoxStream;
 
+use mio::IoVec;
+
 /// A convenience typedef around a `Future` whose error component is `io::Error`
 pub type IoFuture<T> = BoxFuture<T, io::Error>;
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks reasonable to me! I'd only have two points of hesitation:

* Another reexported type from `mio`, but seeing how we've already got one, `EventedFd`, I don't think this is a problem
* Trait methods vs inherent. Eventually we'll probably do away with the `Io` trait in one form or another, as there are quite a lot of I/O conveniences we'd like to have. For now though I think it's ok.

@carllerche @aturon any thoughts on this?

carllerche:
I would like to avoid the bufs suffix. I would like to have read_buf / write_buf be for ops using "bytes" buffers. Maybe _vec instead since that is what readv/writev represents. 

I agree about the mio reexport. I also would like the fn to work with bytes buffers but I don't know how plausible that is.

Io is going away but it will most likely be replaced with some pair of traits for read / write. 

That being said, I think this fn should live on the specific structs that support it for now

seanmonstar:
- The names were to just mirror the names in mio. They can be changed if that's preferable.
- I'm not sure how to make them work with bytes buffers also, that'd depend on them being able to be converted into `IoVec`, wouldn't it?
- I'd strongly push that methods like these exist on the traits some where, otherwise implementations generic of `Io` (or whatever future traits they are) cannot make use of them. I'd prefer to be able to use this in hyper, but if it is only available as inherent methods of `TcpStream`, then I cannot use it.

seanmonstar:
I've updated the names to `read_vec` and `write_vec`, and also updated the default implementation of `write_vec` to not create a copy if only 1 `IoVec` is supplied.

----

Would there be interest in instead of accepting `IoVec`, some sort of opaque struct in tokio-core that can be converted into an `IoVec` internally, and some way for users to convert their arguments into it as well?

alexcrichton:
Yeah eventually we'll probably split `Io` into read/write halves w/ traits, but that can always happen later. Changing the name to have `vec` in the name is fine by me.

@seanmonstar how come the write implementation concatenates but the read one only works on the first buffer? I'd naively expect the write impl to only work on the first buffer as well.

I also figured that conversions could be done via `.into()` to avoid importing a type, perhaps?

seanmonstar:
`read_vec` only reads into one buffer because to do otherwise would invoke multiple syscalls. That seems probably less desirable.

`write_vec` default implementation concatenates into one write because it still is only one syscall, and the performance is usually better than only writing one of the buffers. For example, in hyper, I would much rather concatenate and write the following (in fact, this is done currently since I cannot use writev) then to do two different writes, because fitting as much into one packet is best: `&[b"HTTP/1.1 200 OK\r\nDate: Now\r\nContent-Length: 12\r\n\r\n", b"Hello, World!"]`. If it only wrote the first buffer, I would probably again be restricted from using this in hyper (maybe not, I'd need to weigh performance improvements of TCP versus regressions on anything else).

I just tried changing the method types to receive `&[&[u8]]` and `&mut [&mut [u8]]` as arguments and use `into()`, but there isn't an implementation for a slice of slices, just a single slice into an `IoVec`. I could use unsafe, and transmute them, or safe code would need to create a new vector and copy the `into`s. I'm guessing best is to do the transmute, and upstream that `From` implementation into mio directly?

A side note: why does `read_bufs` take `&mut [&mut [u8]]`? The slice of slices doesn't need to be mutable.

alexcrichton:
If we're going to have default implementations of these functions then I'd kind prefer to only work on the first buffer if multiple are provided because it seems less surprising. Whether that works for hyper's use case though may mean that you want to vendor your own version of this trait.

We must avoid `&[&[u8]]` literally as-is because the representation of `&[u8]` is undefined, but `IoVec` is defined (that was the intention)

seanmonstar:
Then it seems I don't quite understand the comment to use `into()`. `What.into(): Which`?

alexcrichton:
Oh currently there's `From<&[u8]> for &IoVec`

seanmonstar:
There is not, however, `From<&[&[u8]]> for &[&IoVec]`, so either:

- We cannot accept `&[&[u8]` as the argument, since we cannot just transmute it.
- Or we do accept that, and construct a temporary `Vec` to push the `IoVec`s into.

alexcrichton:
Yes that impl intentionally doesn't exist, the thinking is that you'd do:

```rust
stream.write_bufs(&[
    a.into(),
    b.into(),
    c.into(),
    d.into(),
]);
```

seanmonstar:
That makes sense. Still don't see a clear way to change:

```rust
fn write_vec(&mut self, bufs: &[&IoVec]])
```

We can't really unroll that inside the impl for `tokio::net::TcpStream`.

alexcrichton:
Yes I agree in that there's not a lot of changes to make here in terms of signature, just semantics. Do you have thoughts on [my thinking](https://github.com/tokio-rs/tokio-core/pull/157#issuecomment-275201966) for only operating on the first buffer?

seanmonstar:
I still prefer concatenating into a single buffer, but I've changed it to only use the first.

alexcrichton:
Ok!

PR #164, opened 2017-02-08
rename sock -> listener in docs

fortytw2:
Makes it much clearer for someone fairly new to rust and tokio (like me) to follow along when the same named variable aren't totally different things :)


alexcrichton:
Thanks!

PR #166, opened 2017-02-11
Allocation in get_mut() always uses 8 * 1024 bytes

Byron:
The previous implementation would always use the capacity
of the previous buffer, which would effectively prevent it
from ever shrinking.

This also means that protocol with greater variance in
possible frame sizes would likely be heavily over-allocating.
If these implementations use zero-copy, this would imply
that even small frames kept alive by the client would use
large amounts of memory.

The change is motivated by the implementation of the
cassandra-protocol, which allows frames of up to 256MB
in size, which solely depend on the kind of query.

### DIFF
--- a/src/io/frame.rs
+++ b/src/io/frame.rs
@@ -153,7 +153,7 @@ impl EasyBuf {
 
         // If we couldn't get access above then we give ourself a new buffer
         // here.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Looks like tests are failing though?

Byron:
Oh, didn't notice that, thanks:)! The latest change is actually not properly tested, I will do that tomorrow.

alexcrichton:
Thanks!

PR #168, opened 2017-02-13
Implement PartialEq for EasyBuf

Byron:
When implementing `tokio-cassandra`, we noticed that having PartialEq on `EasyBuf` makes testing so much easier, and with that using types that have `EasyBuf` as backing store.


alexcrichton:
Thanks!

PR #169, opened 2017-02-13
add EasyBuf::slice(&self, range: Range<usize>) -> EasyBuf (cf. #120)

manuel-woelker:


alexcrichton:
Thanks for the PR! I'd sort of like to hold off on adding this though until we can get all of the variations to work like `std::slice`, e.g.:

```rust
&buf[..1];
buf.get(2..);
&buf[a..b];
```

tailhook:
Well, it can be done now, by introducing a temporary type, [like here]( https://docs.rs/netbuf/0.3.8/netbuf/struct.Buf.html#method.remove_range). And I think it's forward-compatible: i.e. at any point in time you can make `RangeArgument` and alias of the same from standard library, when that becomes stable. Not sure if it worth it, though.

carllerche:
I agree with @alexcrichton, we should hold off for now.

alexcrichton:
@tailhook yeah that's definitely true, we could vendor our own trait implementation and implement that for everything appropriately. If we feel the need to do so I wouldn't mind merging such a PR!

manuel-woelker:
Thanks for the feedback. Waiting for `RangeArgument` and/or `IndexGet` to stabilize.

PR #173, opened 2017-02-18
Add PartialOrd, Ord, and Hash to EasyBuf, generalize PartialEq.

partim:
This PR adds the derivable traits to `EasyBuf` so that types containing it can have the usual `#[derive()]`s.

It also generalizes `PartialEq` to all `T: AsRef<[u8]>`.


alexcrichton:
Thanks!

PR #174, opened 2017-02-19
Replace magic numbers with constants from mio

asomers:


asomers:
Travis failed due to a failure of rustup, not anything in tokio-core.  Here is the error message:
`/Users/travis/build.sh: line 155: shell_session_update: command not found`

alexcrichton:
Thanks for the PR, but these aren't intended to mirror mio, just be constants here. Perhaps they could be extracted to a private constant in tokio-core though?

asomers:
The Appveyor failure is unrelated to my PR.  The failed test was two_times, which I noticed has also failed on other runs, such as https://ci.appveyor.com/project/carllerche/tokio-core/build/1.0.313

alexcrichton:
Hm these should be used as well when the value is read, right? I forget precisely where that happens but we may not get much benefit from just renaming the store site instead of the load site as well.

PR #178, opened 2017-02-24
Don't call decode if there is no new data.

ahmedcharles:


alexcrichton:
Thanks! Would it be possible to add a test for this?

carllerche:
I'm not sure that this is a correct change. It has previously been acceptable for `Codec` to return frames even if the supplied buffer was empty. I also have previously taken advantage of this feature.

Take the following HTTP sequence:

* HTTP header including content-length: 100
* 100 bytes of body
* HTTP header for next request

I implemented the frames to be:

```
enum HttpFrame {
    Head(...),
   Body(Option<Chunk>),
}
```

The codec would keep local state tracking the length of the streaming body. When decoding the body stream, it would drain from the supplied buffer, then return `Some(chunk)`. If decoding the **last** chunk, it would do: `self.end_of_body_stream = true`

The next call to `decode`, the supplied buffer **could** be empty, but the `decode` fn would see that `end_of_body_stream` was set to `true` and emit a `Body(None)` frame (while unsetting the flag).

This change would break this behavior.

alexcrichton:
@carllerche note though that your example could be changed to `Body { chunk: Chunk, last: bool }`, right? 

I do think it'd be breaking though, which is unfortunate.

alexcrichton:
@ahmedcharles at the very least, it may be prudent to document that this should be handled in `decode` regardless?

carllerche:
@alexcrichton the `Body(Option<Chunk>)` pattern is what tokio-proto uses: https://github.com/tokio-rs/tokio-proto/blob/master/src/streaming/pipeline/frame.rs#L12-L17 It would not be trivial to change it.

alexcrichton:
Oh true yeah, good point

carllerche:
Thanks for submitting the PR. I'm going to close it in favor of [tokio-io](http://github.com/alexcrichton/tokio-io) which is taking over `Io` and `Framed` in tokio-core.

The behavior of calling `decode` when there is no data has been maintained due to the points raised above. Additionally, `decode_eof` is going to be called as well, to match the behavior.

PR #185, opened 2017-03-03
Don't fail UdpFramed due to transient errors

Ralith:
UDP operations have a variety of transient errors under normal operation. `UdpFramed` currently bails out if any are encountered, which is usually not desired. This change exposes `ECONNRESET` on the `Stream` and suppresses `EPERM` on the `Sink`. Writes can also return the nonfatal error `EMSGSIZE`, but I'm not sure if there's a good way to identify that error from inside tokio-core, is not transient in the sense that future transmits of the same message may succeed, and should usually be statically avoidable by not sending unreasonably large datagrams anyway.

This change also introduces an associated error type because decoder errors are not IO errors.

### DIFF
--- a/src/net/udp/frame.rs
+++ b/src/net/udp/frame.rs
@@ -26,6 +26,9 @@ pub trait UdpCodec {
     /// The type of frames to be encoded.
     type Out;
 
+    /// The type of fatal decoding errors.
+    type Error;
+
     /// Attempts to decode a frame from the provided buffer of bytes.
     ///
     /// This method is called by `UdpFramed` on a single datagram which has been
@@ -37,7 +40,7 @@ pub trait UdpCodec {
     /// Finally, if the bytes in the buffer are malformed then an error is
     /// returned indicating why. This informs `Framed` that the stream is now
     /// corrupt and should be terminated.
-    fn decode(&mut self, src: &SocketAddr, buf: &[u8]) -> io::Result<Self::In>;
+    fn decode(&mut self, src: &SocketAddr, buf: &[u8]) -> Result<Self::In, Self::Error>;
 
     /// Encodes a frame into the buffer provided.
     ///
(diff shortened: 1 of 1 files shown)


alexcrichton:
This seems like it's going into the realm of "not easy mode" and perhaps doesn't belong in tokio-core at that point? The encoders/decoders here are just intended to help a project get up and running quickly, not necessary take care of all possible use cases. I'd be pretty wary about indicating that errors should be ignored by default for most projects...

Ralith:
This isn't motivated by a desire to take care of all possible use cases. Both the `EPERM` and `ECONNRESET` situations are IMO easy for *any* UDP-based server to encounter, and handling them by aborting IO on the socket will make most code using the existing `UdpFramed` very fragile. Ignoring errors by default isn't great, I agree, but perhaps it's still better than dying whenever a client closes their socket abruptly or the host gets briefly overloaded? "Easy mode" to me means automatically doing what most applications require, rather than being as lossless as possible at the cost of sometimes surprising behavior.

Maybe `UdpFramed` is fine as it is for the sake of just making it easy for people to prototype, but I'm leery of defaults that lead to unreliable software. Treating inherently transient UDP errors as fatal strikes me as a footgun. Nonetheless, these changes were mostly exploratory; feel free to close if you're certain the current behavior is preferred.

alexcrichton:
At the very least if we are to accept this it has to be documented *somewhere*, and in general I think the same leriness can be applied to discarding errors silently by default. Can this be configured perhaps?

Ralith:
Another option might be to provide combinators to conveniently filter errors in `Sink`s and `Stream`s and recommend their use in the UDP documentation. That might be at odds with a general principle that errors are always fatal, though. If errors at any level of a `Stream`/`Sink` *are* to always be considered fatal, then it's not clear to me that the existing implementation `UdpFramed` is usable at all in robust applications. In particular, I'm not aware of a good way to recover from an error in `UdpFramed` without recreating the UDP socket from scratch.

While I agree that suppressing errors in general is inadvisable, I think suppressing these specific errors is acceptable because the semantics of UDP are such that a well-behaved application will remain well-behaved if these errors are not propagated. By contrast, dying on `ECONNRESET` represents a DoS vulnerability for servers. Configurability seems reasonable (particularly for `ECONNRESET`), though in the interest of minimizing footguns I'm strongly in favor of making the suppression opt-out. If this approach seems palatable to you I'll draft an appropriate patch.

alexcrichton:
I think it's fine to document yeah that various combinators are recommended to be used. I also think it's ok for some sinks/streams to not panic after they generate errors, it's just that if you're working with a generic sink/stream it's all you can assume.

I'd still be wary to change the default behavior, but adding knobs seems ok.

Ralith:
Closing this since it sounds like suppressing errors manually is the way to go. It'd still be useful to document this and add some helpers to `Sink` to make it easier to suppress errors there, but that's sufficiently different in spirit to be done as a separate PR, I think.

e: and involves changes in a different repository entirely, for that matter.

PR #195, opened 2017-03-27
Fix logic error in TcpStream.read_buf()

living180:
If the underlying `read_bufs()` call returned a `WouldBlock` error, `TcpStream.read_buf()` was erroneously calling `self.io.need_write()`, when it should actually call `self.io.need_read()`.


alexcrichton:
Oops looks like https://github.com/tokio-rs/tokio-core/commit/45e69bfb61d5638cd89b8ed0c2808738a9c8802b needs to be for `read_buf` as well!

PR #199, opened 2017-04-25
AIO support

asomers:
Add definitions for asynchronous file I/O.  The mio-aio crate will
provide a concrete implementation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ appveyor = { repository = "alexcrichton/tokio-core" }
 [dependencies]
 bytes = "0.4"
 log = "0.3"
-mio = "0.6.5"
+mio = { git = "https://github.com/asomers/mio", branch = "aio6" }
 scoped-tls = "0.1.0"
 slab = "0.3"
 iovec = "0.1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -19,7 +19,7 @@ appveyor = { repository = "alexcrichton/tokio-core" }
 [dependencies]
 bytes = "0.4"
 log = "0.3"
-mio = "0.6.5"
+mio = { git = "https://github.com/asomers/mio", branch = "aio6" }
```
asomers, line 22: I'll delete this part of the diff as soon as https://github.com/carllerche/mio/pull/586 merges

alexcrichton:
Thanks for the PR! This to me definitely raises the question of how to expose platform-specific events through tokio-core. Ideally we wouldn't have to add a new method for all events that we're listening for over time, as it's tough to do that in a platform-specific fashion.

I wonder if we could perhaps generalize `PollEvented` a bit to work with an instance of `Ready`? That would push that platform-specific logic down to mio and we could maybe refactor the read/write readiness methods to working with this api as well.

alexcrichton:
Ok I've added a generalization of this in https://github.com/tokio-rs/tokio-core/pull/208 which I think is slightly more flexible than adding a method-per-event.

alexcrichton:
Ok I've merged https://github.com/tokio-rs/tokio-core/pull/208, so I'm going to close this in favor of that.

PR #205, opened 2017-05-09
small typo fixed for io module

MarkusJais:
Just a very tiny typo fixed. "objects" should be plural in that sentence"


alexcrichton:
Thanks!

PR #210, opened 2017-05-21
POSIX AIO support, try 2

asomers:
Support POSIX AIO, post 01635df .  A concrete implementation will be
added by the mio-aio and tokio-file crates

### DIFF
--- a/src/reactor/mod.rs
+++ b/src/reactor/mod.rs
@@ -469,8 +469,7 @@ impl Inner {
                               mio::Token(TOKEN_START + entry.index() * 2),
                               mio::Ready::readable() |
                                 mio::Ready::writable() |
-                                platform::hup() |
-                                platform::aio(),
+                                platform::all(),
                               mio::PollOpt::edge()));
         Ok((sched.readiness.clone(), entry.insert(sched).index()))
     }
@@ -748,7 +747,7 @@ impl<F: FnOnce(&Core) + Send + 'static> FnBox for F {
 }
 
 fn read_ready() -> mio::Ready {
-    mio::Ready::readable() | platform::hup() | platform::aio()
+    mio::Ready::readable() | platform::hup()
 }
 
 const READ: usize = 1 << 0;
@@ -786,6 +785,10 @@ mod platform {
         // don't implement it with their reactors, so there's no point
         // to using it.
         Ready::empty()
+    }
+
+    pub fn all() -> Ready {
+        hup()
     }
 
     pub fn hup() -> Ready {
@@ -826,6 +829,10 @@ mod platform {
 
     pub fn aio() -> Ready {
         UnixReady::aio().into()
+    }
+
+    pub fn all() -> Ready {
+        hup() | aio()
     }
 
     pub fn hup() -> Ready {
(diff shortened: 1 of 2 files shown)


src/reactor/mod.rs
```diff
@@ -469,7 +469,8 @@ impl Inner {
                               mio::Token(TOKEN_START + entry.index() * 2),
                               mio::Ready::readable() |
                                 mio::Ready::writable() |
-                                platform::hup(),
+                                platform::hup() |
+                                platform::aio(),
```
alexcrichton: Oh this is something I should have fixed earlier, mind changing this to `platform::all()` and folding both `hup` and `aio` into that one function?

src/reactor/mod.rs
```diff
@@ -747,7 +748,7 @@ impl<F: FnOnce(&Core) + Send + 'static> FnBox for F {
 }
 
 fn read_ready() -> mio::Ready {
-    mio::Ready::readable() | platform::hup()
+    mio::Ready::readable() | platform::hup() | platform::aio()
```
alexcrichton: I think this change may not be necessary? The thinking here for `hup` was that pipes on Unix and such typically use `hup` as "this is now readable at eof" and they may not otherwise generate a readable event (I think there's a test case for this?) but for general I/O objects I don't think `aio` should by in the default read events?

You'd still get notifications, though, through `poll_ready` of course.

alexcrichton (commented):
Looks great to me, thanks!

src/reactor/mod.rs
```diff
@@ -775,21 +775,80 @@ fn usize2ready(bits: usize) -> mio::Ready {
     ready | platform::usize2ready(bits)
 }
 
-#[cfg(unix)]
+#[cfg(all(unix, not(any(target_os = "freebsd", target_os = "dragonfly"))))]
 mod platform {
     use mio::Ready;
     use mio::unix::UnixReady;
 
+    pub fn aio() -> Ready {
```
alexcrichton: I think this function may no longer be needed, right?

src/reactor/mod.rs
```diff
@@ -775,21 +775,80 @@ fn usize2ready(bits: usize) -> mio::Ready {
     ready | platform::usize2ready(bits)
 }
 
-#[cfg(unix)]
+#[cfg(all(unix, not(any(target_os = "freebsd", target_os = "dragonfly"))))]
```
alexcrichton: This `#[cfg]` isn't present in mio, so we shouldn't need it here, right?

src/reactor/mod.rs
asomers: Technically not, but I left it there for symmetry.

src/reactor/mod.rs
asomers: It's true that we could combine the two sections and use the FreeBSD/Dragonfly definitions for all Unixes.  The downside is that platforms that don't support POSIX AIO would have to do slightly more work in `ready2usize` and `usize2ready`.  What's your preference?

alexcrichton: Eh I think it's ok, let's collapse these

src/reactor/mod.rs
alexcrichton: Ah ok, let's remove it as I think it's dead code?

alexcrichton:
Ok looks good, thanks!

PR #223, opened 2017-06-23
Make TimeoutToken public

en:
Just export TimeoutToken.


alexcrichton:
This type is intended to be private and in theory is superseded by the `Interval` and `Timeout` types, how come you need this public?

en:
Hi, Alex! I implemented a custom type similar to `Interval`, but the interval is different in each iteration.

alexcrichton:
Ah ok makes sense to me. Becaues this time is intended to be private, I wonder if we could add an approrpriate method to the `Timeout` type perhaps?

en:
How about add this method to `Timeout`:
```
    pub fn reset(&mut self, at: Instant) {
        self.when = at;
        self.token.reset_timeout(self.when, &self.handle);
    }
```
So I can reuse a `Timeout` to implement a custom `Interval`.

alexcrichton:
Yeah sounds great!

en:
OK, I will submit a new PR.

PR #224, opened 2017-06-23
make TcpStream::flush() a noop

seanmonstar:
`std::net::TcpStream::flush` is at the end of the current call chain, and it does nothing. However, the wrapper in tokio-core does an atomic compare for write readiness before calling in to the noop. This actually shows up in profiles, and since it does nothing in the end, let's skip the atomic operations.


alexcrichton:
👍

PR #229, opened 2017-07-03
Make Error an associated type of UdpCodec

Eroc33:
As has been done in tokio-io with `Encoder` and `Decoder`


alexcrichton:
Thanks for the PR! This is a breaking change so we can't land this just yet but I've tagged it as such for the next release.

PR #237, opened 2017-07-31
[fuchsia] Config changes to build on Fuchsia

raphlinus:
This patch disables various Unix-specific platform features that are not enabled on Fuchsia. It also updates the mio version to 0.6.10, which is the first release that supports Fuchsia.


alexcrichton:
👍

PR #238, opened 2017-08-16
update example in doc comment

cssivision:
some methods used by example in doc comment has deprecated, and move into tokio_io, so update it.


alexcrichton:
Thanks!

PR #244, opened 2017-09-03
Prefer ErrorKind::WouldBlock

### LINKED ISSUE #216: Take advantage of ErrorKind::WouldBlock.into()
Stable since 1.14 and it's probably faster (https://github.com/rust-lang/rust/pull/42426)

arthurprs:
fix #216

This is probably measurable in practice

```
pub fn before() -> io::Error {
    io::Error::new(io::ErrorKind::WouldBlock, "would block")
}

pub fn after() -> io::Error {
    io::ErrorKind::WouldBlock.into()
}

example::before:
        push    rbp
        mov     rbp, rsp
        push    rbx
        push    rax
        mov     rbx, rdi
        lea     rdi, [rip + str.0]
        mov     esi, 11
        call    std::error::<impl core::convert::From<&'b str> for alloc::boxed::Box<std::error::Error + core::marker::Sync + core::marker::Send + 'a>>::from@PLT
        mov     rcx, rdx
        mov     esi, 10
        mov     rdi, rbx
        mov     rdx, rax
        call    std::io::error::Error::_new@PLT
        mov     rax, rbx
        add     rsp, 8
        pop     rbx
        pop     rbp
        ret

example::after:
        push    rbp
        mov     rbp, rsp
        mov     word ptr [rdi], 2561
        mov     rax, rdi
        pop     rbp
        ret

str.0:
        .ascii  "would block"
```

### DIFF
--- a/src/io/split.rs
+++ b/src/io/split.rs
@@ -46,7 +46,7 @@ impl<T: Read> Read for ReadHalf<T> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         match self.handle.poll_lock() {
             Async::Ready(mut l) => l.read(buf),
-            Async::NotReady => Err(::would_block()),
+            Async::NotReady => Err(io::ErrorKind::WouldBlock.into()),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks!

PR #246, opened 2017-09-08
bump slab to 0.4.0

twmb:
Makes some of the code easier to read.

I believe this is all correct, but it's worth a close look.


src/reactor/mod.rs
```diff
@@ -544,18 +547,18 @@ impl Inner {
     fn cancel_timeout(&mut self, token: usize) {
         debug!("cancel a timeout: {}", token);
         let pair = self.timeouts.remove(token);
-        if let Some((Some(slot), _state)) = pair {
+        if let (Some(slot), _state) = pair {
```
twmb, line 550: If `cancel_timeout` could be called twice for the same token in the past, this change is invalid. 0.3.0 would return `None` if the token did not exist; 0.4.0 will panic. I haven't vetted if `cancel_timeout` can be called twice.

alexcrichton:
Looks great, thanks!

PR #247, opened 2017-09-09
TokenTimeout: add doc around new panic

twmb:
My slab bump in #246 added a potential new panic in Core's Inner's
`cancel_timeout`.

Thankfully, the only way `cancel_timeout` can be called is from
TimeoutToken's `cancel_timeout`, which is crate-internal only and is
only called from Timeout's and Interval's drop fn's.

This change simply adds new clarifying documentation around
TokenTimeout's cancel_timeout to "future proof" anybody looking to use
cancel_timeout directly (not just on drop).

### DIFF
--- a/src/reactor/timeout_token.rs
+++ b/src/reactor/timeout_token.rs
@@ -48,7 +48,7 @@ impl TimeoutToken {
     /// # Panics
     ///
     /// This method will panic if the timeout specified was not created by this
-    /// loop handle's `add_timeout` method.
+    /// loop handle's `add_timeout` method or if called multiple times.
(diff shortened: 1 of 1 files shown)


twmb:
It may also be worth it to mention that `cancel_timeout` is called automatically in the `Timeout` and `Interval` docs, or to just change the core's `Inner` struct to [check](https://docs.rs/slab/0.4.0/slab/struct.Slab.html#method.contains) first before removing from the slab.

alexcrichton:
Thanks! This is an internal type not intended for public consumption, and usage of `Interval` or `Timeout` should never cause a panic, but it's fine to document the internals here.

PR #250, opened 2017-09-11
UDP 0-length datagrams

### LINKED ISSUE #248: UdpFramed doesn’t send zero length UDP packets
[`UdpFramed` doesn’t call `send_to()`](https://github.com/tokio-rs/tokio-core/blob/645ae7051d08221cbd44ad1211eac8af2f0b6c67/src/net/udp/frame.rs#L79-L113) on the underlining `UdpSocket` if no data is written to the send buffer by the `UdpCodec`. `UdpSocket` sends zero length packets as expected.

henninglive:
Resolves #248, preventing UdpFramed from sending 0-length datagrams.
Also, adds 0-length tests for UdpSocket.

### DIFF
--- a/src/net/udp/frame.rs
+++ b/src/net/udp/frame.rs
@@ -61,6 +61,7 @@ pub struct UdpFramed<C> {
     rd: Vec<u8>,
     wr: Vec<u8>,
     out_addr: SocketAddr,
+    flushed: bool,
 }
 
 impl<C: UdpCodec> Stream for UdpFramed<C> {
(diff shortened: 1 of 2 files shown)


src/net/udp/frame.rs
```diff
@@ -81,29 +82,32 @@ impl<C: UdpCodec> Sink for UdpFramed<C> {
     type SinkError = io::Error;
 
     fn start_send(&mut self, item: C::Out) -> StartSend<C::Out, io::Error> {
-        if self.wr.len() > 0 {
+        trace!("sending frame");
+
+        if !self.flushed {
             try!(self.poll_complete());
-            if self.wr.len() > 0 {
```
alexcrichton: I think this clause may still be needed? If `poll_complete` returns `NotReady` then we need to return `NotReady` here I think?

alexcrichton:
Thanks! Other than one small nit looks good to me!

henninglive:
Fixed

alexcrichton:
Hm maybe a forgotten push? I think the clause is still missing?

henninglive:
yeah, sorry. Pushed now.

alexcrichton:
Thanks!

PR #251, opened 2017-09-11
Adds UdpSocket connect, send and recv methods

bkchr:


src/net/udp/mod.rs
```diff
@@ -74,6 +74,24 @@ impl UdpSocket {
...
+    }
+
+    /// Sends data on the socket to the address previously bound via connect(). 
+    /// On success, returns the number of bytes written.
+    pub fn send(&self, buf: &[u8]) -> io::Result<usize> {
+        self.io.get_ref().send(buf)
+    }
+
+    /// Receives data from the socket previously bound with connect(). 
+    /// On success, returns the number of bytes read and the address from whence the data came.
```
alexcrichton: This isn't quite right I think? (no address is returned)

examples/echo-udp.rs
```diff
@@ -44,7 +44,7 @@ impl Future for Server {
 
             // If we're here then `to_send` is `None`, so we take a look for the
             // next message we're going to echo back.
-            self.to_send = Some(try_nb!(self.socket.recv_from(&mut self.buf)));
+            self.to_send = Some(match self.socket.recv_from(&mut self.buf) { Ok(res) => res, e@_ => { println!("{:?}", e); return Ok(::futures::Async::NotReady); }});
```
alexcrichton: stray change?

alexcrichton:
Thanks! Could you add a test for this as well?

examples/echo-udp.rs
bkchr: Ohh. That was some test code..

bkchr:
Okay, I addressed your comments and added a test :)

alexcrichton:
Thanks!

PR #270, opened 2017-10-26
Fix warnings

cramertj:


src/lib.rs
```diff
@@ -91,6 +91,8 @@
 
 #![doc(html_root_url = "https://docs.rs/tokio-core/0.1")]
 #![deny(missing_docs)]
+#![deny(warnings)]
+#![cfg_attr(test, allow(deprecated))]
```
cramertj, line 95: I'm not totally sure why this line is necessary-- without it, `cargo test` fails with "use of deprecated item: moved to the `tokio-io` crate", but it doesn't point to any actual code.

src/lib.rs
```diff
@@ -91,6 +91,8 @@
 
 #![doc(html_root_url = "https://docs.rs/tokio-core/0.1")]
 #![deny(missing_docs)]
+#![deny(warnings)]
```
cramertj, line 94: Up to you whether to include this or not.

alexcrichton:
Thanks!

PR #271, opened 2017-10-26
Simplify UdpSocket futures

cramertj:
This is just some cleanup that makes the code easier to read for me personally. Feel free to close this if you think it isn't an improvement. I'd been pointing some people at this code as an example of how to write non-blocking futures with tokio-core, and they found the old code hard to parse.


alexcrichton:
Thanks! I think I may have accidentally merged in the wrong order creating conflicts though, want to rebase?

cramertj:
Nah, no problem-- I don't think there was any order in which you could have merged them that wouldn't have caused conflicts.

PR #281, opened 2017-12-01
Add Timeout::unregister

### LINKED ISSUE #280: Allow timers to be reset without a new deadline
I've recently refactored tokio-io-timeout to reuse the same `Timeout` rather than repeatedly creating and destroying them. When I need to "turn off" the timeout, I can use reset, but that requires me to provide a new deadline. It'd be nice if there was a variant that didn't, and simply unregistered the timeout from the timer heap.

https://github.com/sfackler/tokio-io-timeout/blob/master/src/lib.rs#L48

sfackler:
This is useful if you are keeping a persistent `Timeout` object which
may or may not be active at any time.

Closes #280 

r? @alexcrichton


src/reactor/mod.rs
```diff
@@ -544,6 +548,15 @@ impl Inner {
         debug!("set a timeout: {}", token);
     }
 
+    fn unregister_timeout(&mut self, token: usize) {
+        let pair = &mut self.timeouts[token];
+        if let Some(slot) = pair.0.take() {
+            self.timer_heap.remove(slot);
+        }
+        pair.1 = TimeoutState::NotFired;
```
sfackler, line 556: Should this be modified? `update_timeout` doesn't touch it so it seems like we might want to leave it alone.

alexcrichton:
Thanks! So reading over the docs/implementation here the purpose is to basically avoid a task notification, right? Out of curiosity, does that bring perf improvements locally? Or is this done mostly out of "seems like the right thing to do"?

sfackler:
We avoid the task notification already by using `reset` - that's important since an extra notification will cause a new timeout to start up without a read/write call actually being made.

I kind of doubt that there'd be a huge performance difference, especially in a traditional IO bound workflow. This is mostly just a "right thing to do" kind of PR - I don't *want* to reschedule the timeout, but just deregister the task from it.

carllerche:
Given that the timer in tokio-core is deprecated as it is being moved out of the crate, should we hold off this PR?

sfackler:
Sure - is the new implementation in a place that I should start poking at/using or is that a tokio 0.2 thing?

carllerche:
The impl was extracted to: http://github.com/alexcrichton/futures-timer

alexcrichton:
@sfackler ok thanks for the info! If this is needed for performance we can merge/release this, but yeah otherwise it may be best to hold off for a bit (although a PR to `futures-timer` would also be welcome!)

sfackler:
Cool, I'll close this and poke at futures-timer.

PR #282, opened 2017-12-01
Add Fuchsia-specific reactor and net impls

cramertj:
This change splits out the reactor into separate mio-based and Fuchsia-ports-based implementations. This is greatly beneficial to Fuchsia's uses of tokio-core, as it allows us to schedule and receive custom notifications via the `PacketReceiver` interface. After this change, the Fuchsia backend for mio will no longer be used by tokio-core.

I apologize this is so much code-- I wanted to do this all at once so that I could see how all the different pieces came together. I hope that the Fuchsia components are isolated enough that it doesn't impose too great of a maintenance burden.

cc @raggi, @raphlinus, @alexcrichton, @carllerche

### DIFF
--- a/src/net/fuchsia/mod.rs
+++ b/src/net/fuchsia/mod.rs
@@ -1,0 +1,369 @@
+mod tcp;
+pub(crate) use self::tcp::{TcpListener, TcpStream};
+
+mod udp;
+pub(crate) use self::udp::UdpSocket;
+
+use futures::Async;
+use futures::task::AtomicTask;
+use libc;
+use zx::{self, AsHandleRef};
+
+use std::io::{self, Read, Write};
+use std::mem;
+use std::net::{self, SocketAddr};
+use std::os::unix::io::{AsRawFd, RawFd};
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+
+use reactor::{self, PacketReceiver, ReceiverRegistration};
+use is_wouldblock;
+
+const READABLE: usize = libc::EPOLLIN as usize;
+const WRITABLE: usize = libc::EPOLLOUT as usize;
+const ERROR: usize    = libc::EPOLLERR as usize;
+const HUP: usize      = libc::EPOLLHUP as usize;
+
+// Unsafe to use. `receive_packet` must not be called after
+// `fdio` is invalidated.
+pub(crate) struct EventedFdPacketReceiver {
+    fdio: *const syscall::fdio_t,
+    signals: AtomicUsize,
+    read_task: AtomicTask,
+    write_task: AtomicTask,
+}
+
+// Needed because of the fdio pointer.
+// It is safe to send because the `EventedFdPacketReceiver` must be
+// deregistered (and therefore `receive_packet` never called again)
+// before `__fdio_release` is called.
+unsafe impl Send for EventedFdPacketReceiver {}
+unsafe impl Sync for EventedFdPacketReceiver {}
+
+impl PacketReceiver for EventedFdPacketReceiver {
+    fn receive_packet(&self, packet: zx::Packet) {
+        let observed_signals = if let zx::PacketContents::SignalOne(p) = packet.contents() {
+            p.observed()
+        } else { return };
+
+        let mut events: u32 = 0;
+        unsafe {
+            syscall::__fdio_wait_end(self.fdio, observed_signals.bits(), &mut events);
+        }
+        let events = events as usize;
+
+        let old = self.signals.fetch_or(events, Ordering::SeqCst);
+        let became_readable = ((events & READABLE) != 0) && ((old & READABLE) == 0);
+        let became_writable = ((events & WRITABLE) != 0) && ((old & WRITABLE) == 0);
+        let err_occurred = (events &(ERROR | HUP)) != 0;
+
+        if became_readable || err_occurred {
+            self.read_task.notify();
+        }
+        if became_writable || err_occurred {
+            self.write_task.notify();
+        }
+    }
+}
+
+/// A type which can be used for receiving IO events for a file descriptor.
+pub(crate) struct EventedFd<T> {
+    inner: T,
+    // Must be valid, acquired from `__fdio_to_io`
+    fdio: *const syscall::fdio_t,
+    // Must be dropped before `__fdio_release` is called
+    signal_receiver: mem::ManuallyDrop<ReceiverRegistration<EventedFdPacketReceiver>>,
+}
+
+unsafe impl<T> Send for EventedFd<T> where T: Send {}
+unsafe impl<T> Sync for EventedFd<T> where T: Sync {}
+
+impl<T> Drop for EventedFd<T> {
+    fn drop(&mut self) {
+        unsafe {
+            // Drop the receiver so `packet_receive` may not be called again.
+            mem::ManuallyDrop::drop(&mut self.signal_receiver);
+
+            // Release the fdio
+            syscall::__fdio_release(self.fdio);
+        }
+
+        // Then `inner` gets dropped
+    }
+}
+
+impl<T> EventedFd<T> where T: AsRawFd {
+    /// Creates a new EventedFd.
+    ///
+    /// For this function to be safe, the underlying file descriptor from `T::as_raw_fd`
+    /// must be a valid file descriptor which remains valid for the duration of `T`'s
+    /// lifetime.
+    pub unsafe fn new(inner: T, remote: &reactor::Remote) -> io::Result<Self> {
+        let fdio = syscall::__fdio_fd_to_io(inner.as_raw_fd());
+        let signal_receiver = remote.register_packet_receiver(Arc::new(EventedFdPacketReceiver {
+            fdio,
+            signals: AtomicUsize::new(0),
+            read_task: AtomicTask::new(),
+            write_task: AtomicTask::new(),
+        }));
+
+        let evented_fd = EventedFd {
+            inner,
+            fdio,
+            signal_receiver: mem::ManuallyDrop::new(signal_receiver),
+        };
+
+        // Make sure a packet is delivered if an error or closure occurs.
+        evented_fd.schedule_packet(ERROR | HUP);
+
+        // Need to schedule packets to maintain the invariant that
+        // if !READABLE or !WRITABLE a packet has been scheduled.
+        evented_fd.schedule_packet(READABLE);
+        evented_fd.schedule_packet(WRITABLE);
+
+        Ok(evented_fd)
+    }
+    /// Tests to see if this resource is ready to be read from.
+    /// If it is not, it arranges for the current task to receive a notification
+    /// when a "writable" signal arrives.
+    pub fn poll_read(&self) -> Async<()> {
+        let receiver = self.signal_receiver.receiver();
+        if (receiver.signals.load(Ordering::SeqCst) &
+            (READABLE | ERROR | HUP)) != 0
+        {
+            Async::Ready(())
+        } else {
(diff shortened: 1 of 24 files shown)


src/reactor/mod.rs
```diff
@@ -4,40 +4,37 @@
...
+use self::fuchsia as sys;
+#[cfg(target_os = "fuchsia")]
+use std::sync::Arc;
 
-mod poll_evented;
 mod timeout;
 mod interval;
-pub use self::poll_evented::PollEvented;
+
+#[cfg(not(target_os = "fuchsia"))]
```
cramertj, line 29: This is the main change to the external interface-- Fuchsia will no longer offer `PollEvented`, but instead exposes the `PacketReceiver` trait which can be registered via a `Core`, `Handle` or `Remote`. After registration, an async message can be queued on the `zx::Port` with the key of the appropriate `ReceiverRegistration`, and the packet will be delivered. This allows for delivery of non-signal-type packets.

src/net/fuchsia/tcp.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn read(&self, buf: &mut [u8]) -> io::Result<usize> {
+        (&self.0).read(buf)
+    }
+
+    pub fn write(&self, buf: &[u8]) -> io::Result<usize> {
+        (&self.0).write(buf)
+    }
+
+    pub fn read_buf<B: BufMut>(&self, buf: &mut B) -> Poll<usize, io::Error> {
+        match (&self.0).read(unsafe { buf.bytes_mut() }) {
```
cramertj, line 139: I didn't do the `IoVec`/`readv` dance here. If I should, let me know-- I wasn't able to find any non-default implementations of `bytes_vec_mut`. If no one is using separate buffers to back a `BufMut`, using `readv` here is a pessimization AFAICT.

src/net/fuchsia/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+            let res = handle.wait_async_handle(
+                self.signal_receiver.remote().port(),
+                self.signal_receiver.key(),
+                signals,
+                zx::WaitAsyncOpts::Once,
+            );
+
+            // The handle is borrowed, so we cannot drop it.
+            mem::forget(handle);
+            res.expect("Error scheduling EventedFd notification");
```
cramertj, line 214: Note: this function cannot return `Result` because `need_read` and `need_write` are infallible.

carllerche:
If [this](https://github.com/carllerche/mio/issues/759) mio issue is resolved, is this PR required?

raphlinus (approved):
This was a large enough diff I wouldn't be able to fully understand it without spending lots more time digging in, but the approach seems valid and I didn't find any problems.

cramertj:
@carllerche Theoretically we could solve some of the same problems by allowing arbitrary signal data. However, I think this approach allows for a much more flexible and performant implementation, and will require less churn in the future. It allows other libraries to declare their own `PacketReceiver` types which understand how to handle incoming readiness/signal data. Without this change, that code would have to be baked into mio.

This change also opens the door for Fuchsia-specific optimizations of reactor primitives, such as the `Timer`/`Interval` implementation here. I'm also interested in adding multithreading support for the Fuchsia reactor as we move over into the new "tokio" crate.

Finally, this PR makes it much simpler to debug and diagnose issues in Fuchsia's tokio support. There are now far fewer layers to dig through when searching for bugs, and the `PacketReceiver` structure makes it easier to track down missed or excess wakeups.

carllerche:
I'm pretty hesitant at this point to start adding significant platform specific code to `tokio` and would rather consider alternate strategies.

aturon:
@carllerche would you be open to having a separate Fuchsia "back end" playing a role similar to mio for Tokio, with most of the code living there? (NB I haven't reviewed this patch carefully so this may not make sense.)

cramertj:
@carllerche Can you say more about your specific concerns? I'm not expecting the Tokio authors to maintain this code, or ever make sure it compiles. This structure opens up a lot of flexibility for us in terms of how we can use Tokio.

If you're just worried about the code structure being muddled by shifting things into mio-specific and non-mio-specific directories, I'm happy to restructure things as you'd prefer. For that matter, we could even have a completely separate `tokio-fuchsia` crate which tokio-core `pub use`s and cfgs out all of its own modules. I'd prefer not to do that as it will make my own life slightly more difficult, but I'm greatly invested in this plan and would like to find a way to make it work.

carllerche:
@cramertj 

This is a pretty big PR w/ not much accompanied documentation. It doesn't explain what `PacketReceiver` is in this case and what the benefits are.

> This change also opens the door for Fuchsia-specific optimizations of reactor primitives, such as the Timer/Interval implementation here. 

Timer / Interval are being removed from Tokio.

> Can you say more about your specific concerns? 

My main concern is that I just don't see the point in having this in Tokio, which aims to be portable. This seems like it should live in a completely different library.

carllerche:
@aturon Tokio is converging to be a light futures + mio shim. The amount of work needed to make such a change makes me think it belongs in a separate library.

carllerche:
@cramertj

And yes, there are maintenance concerns. Unlike Mio, which has a very limited integration surface, this would almost definitely break a lot more often.

cramertj:
> @aturon Tokio is converging to be a light futures + mio shim. The amount of work needed to make such a change makes me think it belongs in a separate library.

FWIW I suspect that the amount of code will be much smaller in the new tokio library, since the overall structure of that reactor is much more similar to the Fuchsia one I have here.

cramertj:
> This is a pretty big PR w/ not much accompanied documentation. It doesn't explain what PacketReceiver is in this case and what the benefits are.

There is fairly extensive documentation for the `PacketReceiver` trait [here](https://github.com/tokio-rs/tokio-core/pull/282/files#diff-5786b2aeccdbb2c67703139f466078a5R102).

carllerche:
@cramertj why can't this be part of mio?

aturon:
@carllerche The problem with not integrating with Tokio is that it becomes difficult to integrate with any part of the async ecosystem that assumes Tokio (rather than just `AsyncRead` etc). Ideally we would continue to be able to offer a portable API to non-mainstream platforms like Fuchsia.

FWIW, `std` is set up in a similar manner, with the surface-level APIs hooking into `sys` modules that are each expected to export a common API. We've allowed non-1st-tier platforms to include their own sys modules, but do not gate on them in CI, which means there's effectively no maintenance burden on the Rust team (we ask the Redox folks to do their own reviews as well).

If plausible, it does seem good to move as much of the support code as possible into a separate crate, like I said above -- that way Tokio itself is effectively able to work with multiple back-ends, each of which is a distinct crate.

carllerche:
@aturon 

> The problem with not integrating with Tokio is that it becomes difficult to integrate with any part of the async ecosystem that assumes Tokio

This PR adds non-portable extensions. Tokio is only providing tcp & udp types. AFAIK fuschia already provides tcp / udp w/ tokio.

carllerche:
@aturon it would probably be helpful to focus on specifics. We both agree that "tokio should work on fuschia".

cramertj:
@carllerche If you're interested in pulling `PacketReceiver` into mio and reimplementing `PollEvented` in and `IoToken` in terms of the `PacketReceiver` structure, I bet we could figure something out. However, that seems like a much bigger change to the existing `mio`/`tokio` interaction surface than the one I'm proposing in this PR. That change would be predicated on moving to the tokio crate since the `PacketReceiver` interface requires access to the `scheduled_io` `Slab` in order to deregister.

aturon:
@cramertj is it possible to provide a non-portable small hook, akin to `Evented`, that would allow incorporating other non-portable functionality through a separate add-on crate?

cramertj:
@aturon 
> @cramertj is it possible to provide a non-portable small hook, akin to Evented, that would allow incorporating other non-portable functionality through a separate add-on crate?

Sure, but you also have to remember all of the other parts of tokio-core that currently know about mio, including all of the `net` impls, the reactor awakeners and readiness indicators, etc. The API surface there is relatively large, and all of that would have to be stubbed out into a trait somehow.

cramertj:
@aturon @carllerche Should we move this chat to Gitter or IRC (or vidyo/hangouts) so that we can have a bit more bandwidth for communicating?

carllerche:
@cramertj 

Can you step back and explain, within the context of *only* `TcpStream`, what the problem is with the current setup.

cramertj:
@carllerche `TcpStream` isn't really an issue-- `TcpStream` is mostly fine the way it is (modulo the annoying bit about it not being deregistered and leaking, but that could be fixed via the new tokio trait). The real win here is getting direct access to the packets coming from `zx::Port` and being able to preprocess them as part of the reactor loop, rather than trying to bubble them through some kind of `mio` readiness type, which will be very challenging since they can contain things like pointers into data buffers and other non-signal information.

carllerche:
@cramertj So why don't you add some way to get the fuschia selector impl out of `Poll` and then let users pass in `PacketReceiver` impls there?

cramertj:
@carllerche 
> @cramertj So why don't you add some way to get the fuschia selector impl out of Poll and then let users pass in PacketReceiver impls there?

In order for that to work, I'd have to have a custom reactor built into the Fuchsia reactor inside of mio which stored all of the `Arc<PacketReceiver>`s, and bubble access to that reactor up through `tokio-core::Handle`.  When polling the reactor, I'd have to poll my internal multiple times, dispatching to the `PacketReceiver`s and only returning when a token meant for a `tokio-core` task or `PollEvented` is received. On top of all that, we'd still have to deal with the extra layer of abstraction `mio` introduces, and the performance penalty of `mio`'s reimplementaiton of features that Fuchsia's reactor already provides. None of the additional reactor facilities mio provides are necessary for our reactor-- they already exist and are implemented natively. Plus, that still wouldn't solve the deregistration problem of `PollEvented`s, and wouldn't allow for us to make platform-specific optimizations to the tokio-core reactor (such as multi-threading).

carllerche:
> In order for that to work, I'd have to have a custom reactor built into the Fuchsia reactor inside of mio

This doesn't sound bad (a "reactor" is only a loop).

> which stored all of the `Arc<PacketReceiver>`s

I'm not sure why this would be required when otherwise not.

> and bubble access to that reactor up through `tokio-core::Handle`

This is already done via `PollEvented` (currently public) and something like `IoToken` could probably be made public too.

> When polling the reactor, I'd have to poll my internal multiple times, dispatching to the PacketReceivers and only returning when a token meant for a tokio-core task or PollEvented is received.

I don't see why this is true. All the work would happen in your fuschia specific reactor and the mio interface is via an `Events`, which is a batch interface. Your description sounds like you need to do one off work.

> On top of all that, we'd still have to deal with the extra layer of abstraction `mio` introduces

What I am suggesting bypasses mio for all non portable fuschia behavior, so afaik you wouldn't pay any additional cost.

> None of the additional reactor facilities mio provides are necessary for our reactor-- they already exist and are implemented natively

Could you be specific? I am not familiar with fuschia, so statements like this doesn't help me understand.

> Plus, that still wouldn't solve the deregistration problem of `PollEvented`s

Could you refresh my memory what the issue is?

> nd wouldn't allow for us to make platform-specific optimizations to the tokio-core reactor (such as multi-threading)

Supporting multi-threading is generally a mio TODO and should be tackled separately, I'm not sure what other platform-specific optimizations you would not be able to do, again, please be specific.

carllerche:
Something that may not be obvious. In this path that I am suggesting, you would end up *not* returning mio Events for resources that go via your `PacketReceiver` interface. You would pull the task and notify it yourself.

cramertj:
Yes, I understand that. I'll respond in more detail tomorrow.

PR #283, opened 2017-12-07
Return progress indication from `core.turn`

tiras-j:
Fairly simple change. I have a use-case where the event loop is being driven externally, across FFI (think libcurl multi interface style) and it would be really nice to be able to crank the event loop through multiple task dispatches that aren't waiting on I/O.

Of course it's possible I'm horribly abusing the library or that there are other mechanisms for observing event progress, so I'd appreciate any advice on that front.

### DIFF
--- a/src/reactor/mod.rs
+++ b/src/reactor/mod.rs
@@ -238,7 +238,7 @@ impl Core {
                     return Ok(e)
                 }
             }
-            future_fired = self.poll(None);
+            future_fired = self.poll(None).0;
         }
     }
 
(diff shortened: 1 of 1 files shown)


carllerche:
While subtle, I'm pretty sure this is actually a breaking change. The next breaking release will be the result of the Tokio reform RFC. I would suggest you check if the proposed changes satisfy your needs.

Specifically, this PR addresses the return value of `turn`: https://github.com/tokio-rs/tokio/pull/60

tiras-j:
Yep that `Turn` type is exactly what I'm looking for, once it does get populated with information about that iteration of the loop of course. Definitely better future-proofing than a raw `bool` type!

Closing for now, looking forward to that reform in the future.

PR #293, opened 2018-01-12
Fix UdpCodec::encode

kpp:
Change the signature of UdpCodec::encode
 
Now it is:
    
```
fn encode(&mut self, msg: Self::Out, buf: &mut Vec<u8>)
                                              -> Result<SocketAddr, Self::Error>;
```

Closes https://github.com/tokio-rs/tokio/issues/79


carllerche:
I would suggest opening this on the tokio repo (not tokio-core) as this is a breaking change and cannot land in tokio-core 0.1.

kpp:
@carllerche OK. But what's the difference between tokio and tokio_core?

carllerche:
@kpp the `tokio` crate is basically going to be `tokio-core` 0.2.

kpp:
So do I have to move from `tokio_core` to `tokio`?

kpp:
> I would suggest opening this on the tokio repo (not tokio-core) as this is a breaking change and cannot land in tokio-core 0.1.

@carllerche here it is https://github.com/tokio-rs/tokio/pull/85

alexcrichton:
Ok I'm gonna close this in favor of https://github.com/tokio-rs/tokio/pull/85, thanks for the PRs @kpp!

PR #295, opened 2018-01-25
use 64 iovecs in TcpStream::write_buf

seanmonstar:
Some background: while switching to `write_buf` in hyper, I noticed the 16 pipelined requests benchmark suffered since the responses where split of 32 buffers (1 headers + 1 body * 16 responses), and so each syscall was only writing half as many responses as when hyper flattens all into a single buffer first.

The easiest knee-jerk reaction is to increase the amount of `IoVec`s used in `TcpStream::write_buf`. I arbitrarily picked 64. The man page for `writev(2)` mentions that the default max in Linux is 1024...

A longer term fix could be to add an associated constant to `Buf`, to allow an implementer to state how many `IoVec`s they expect to need.


carllerche (approved):
LGTM, waiting for CI to pass.

PR #302, opened 2018-02-12
lio_listio support

asomers:
This adds support for lio_listio(2), a way of submitting multiple aio_read(2), aio_write(2), and/or aio_fsync(2) operations in a single system call.  Higher level support will be provided by the tokio-file crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ appveyor = { repository = "alexcrichton/tokio-core" }
 [dependencies]
 bytes = "0.4"
 log = "0.4"
-mio = { git = "https://github.com/carllerche/mio", rev = "726b1f3078e79e69c3961f8af32e76e06c49a3f2" }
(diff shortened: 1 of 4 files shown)


carllerche:
Thanks,

Could this be done on the [tokio](http://github.com/tokio-rs/tokio) repo? I will be updating tokio-core to use that crate internally shortly.

asomers:
What's the difference between tokio and tokio-core?  Is the latter being split up into multiple smaller crates or something?

carllerche:
`tokio-core` is essentially deprecated.

carllerche:
I added a deprecation notice to the README: https://github.com/tokio-rs/tokio-core/commit/f765c8f59760b8e899c81a649f5d70f9f7c50f6e

asomers:
Merged into tokio instead of tokio-core .

PR #312, opened 2018-03-11
Drain remote queue before blocking.

### LINKED ISSUE #311: Remote-related bug in tokio-core >= 0.1.13
Hi!

It appears that the new versions of `tokio-core` based on `tokio` (`tokio_core` >= 0.1.13) have broken Rusoto (https://github.com/rusoto/rusoto/issues/1000).

I was able to reduce it to a simple example program that works correctly with `tokio_core` 0.1.12 (`tokio_io` 0.1.5), but breaks with `tokio_core` 0.1.13 or 0.1.14.

Here is the program:

```rust
extern crate tokio_core;

use std::thread;
use std::sync::{Arc, Barrier, Mutex};

use tokio_core::reactor::{Core, Remote};

fn spawn() -> Remote {
    let slot = Arc::new(Mutex::new(None));
    let slot_clone = slot.clone();

    thread::spawn(move || {
        let mut core = Core::new().unwrap();
        let mut guard = slot_clone.lock().unwrap();
        *guard = Some(core.remote());
        drop(guard);
        println!("turning core...");
        loop {
            core.turn(None);
        }
    });

    loop {
        let mut guard = slot.lock().unwrap();
        if let Some(remote) = guard.take() {
            return remote;
        }
    }
}

fn main() {
    let remote = spawn();
    let barrier = Arc::new(Barrier::new(2));
    let barrier_clone = barrier.clone();

    remote.spawn(move |_| {
        println!("waiting in remote");
        barrier_clone.wait();
        println!("waking in remote");
        Ok(())
    });

    println!("waiting in main");
    barrier.wait();
    println!("waking in main");
}
```

The expected behaviour of the program is to print the following and then exit:

```
turning core...
waiting in main
waiting in remote
waking in remote
waking in main
```

However, using `tokio_core` 0.1.13 or 0.1.14, the output is the following, and it never exits:

```
turning core...
waiting in main
```

Superficially it looks like the closure passed to `remote.spawn` is never executed.

carllerche:
Unless this is done before blocking, sending a command over the remote
channel will not be able to wake up a blocked reactor.

Fixes #311

### DIFF
--- a/src/reactor/mod.rs
+++ b/src/reactor/mod.rs
@@ -287,6 +287,11 @@ impl Core {
                     (max_wait, timeout) => max_wait.or(timeout),
                 };
 
+                // Process all the events that came in, dispatching appropriately
+                if self.notify_rx.take() {
+                    CURRENT_LOOP.set(self, || self.consume_queue());
+                }
+
                 // Drain any futures pending spawn
                 {
                     let mut e = self.executor.borrow_mut();
@@ -315,11 +320,6 @@ impl Core {
                 // Process all timeouts that may have just occurred, updating the
                 // current time since
                 self.consume_timeouts(after_poll);
-
-                // Process all the events that came in, dispatching appropriately
-                if self.notify_rx.take() {
-                    CURRENT_LOOP.set(self, || self.consume_queue());
-                }
 
                 debug!("loop process, {:?}", after_poll.elapsed());
             });


srijs:
Thanks @carllerche! Can you advise on whether we should try to deploy a workaround in Rusoto until this lands in a new tokio-core release? Or is the plan to do a patch release soon-ish?

carllerche:
@srijs i'll push it tonight. There are a couple of other issues I want to try to solve too.

carllerche:
The fix is out.

PR #320, opened 2018-04-10
Integrate with tokio-timer

srijs:
This PR includes two commits that do two things:

1. Integrate `Core` with `tokio-timer`, so that functionality that relies on a default timer works in the "compatibility mode" of using `tokio-core`. See https://github.com/srijs/rust-tokio-retry/issues/8 for a use-case.

2. Replace the internal timer heap with the implementation in `tokio-timer`, removing a bunch of code.

I'm aware that `tokio-core` is in "maintenance" mode at this point, but I feel that (1) is important enough to make it into a release so that the ecosystem (such as `tokio-retry`) can fully switch to using the new tokio/tokio-timer while still maintaining compatibility with `tokio-core` (which was the original intention behind re-basing `tokio-core` on `tokio`, iiuc).

I feel less strongly about (2), just thought it was a nice to be able to remove some things that essentially seemed redundant at this point. Happy to revert that commit if you don't think it's worth it.

Let me know what you think!


carllerche:
This is great, I’ve been meaning to do this. I’ll review in a bit. 

Even though it is in “maintenance mode”, I think it is good to try to make it as easy as possible to upgrade.

carllerche:
Looks great. Thanks!

PR #325, opened 2018-05-21
Remove references to "LoopData", "lp" and improve Core::run in docs

cpick:
Remove reference to obsolete `LoopData`.
Rename `lp` variable in example to `core` to match current naming conventions.

Try to clarify `Core::run`s docs to indicate that it only waits for its
argument's future to complete before returning.


carllerche:
Thanks!

